PR #11, opened 2020-08-26
Enable PR review requirement

cevich:

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -15,8 +15,7 @@ pull_request_rules:
           - '-title~=.*CI.*SKIP'  # Cirrus-CI feature
           - '-title~=.*SKIP.*CI'  # "
           - 'status-success=success'  # defined in .cirrus.yml
-          # Temporarily disabled during repository bootstrapping
-          # - '#approved-reviews-by>=1'
+          - '#approved-reviews-by>=1'
       actions:
           merge:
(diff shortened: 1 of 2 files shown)


cevich:
@edsantiago if you don't mind.  Please review and mark 'approved'.  I want to test/make sure the automatic merge system works.

.mergify.yml
```diff
@@ -15,8 +15,7 @@ pull_request_rules:
           - '-title~=.*CI.*SKIP'  # Cirrus-CI feature
           - '-title~=.*SKIP.*CI'  # "
           - 'status-success=success'  # defined in .cirrus.yml
-          # Temporarily disabled during repository bootstrapping
-          # - '#approved-reviews-by>=1'
+          - '#approved-reviews-by>=1'
```
edsantiago, line 18: I have to assume that the `#` here, inside a string, is an actual part of the conditional. Is this under your control? Is there any possible way to remove the hash sign? It's super confusing. But I'll approve anyway.

cevich: confirmed, here's the spot in docs: https://doc.mergify.io/conditions.html#conditions

cevich:
Ahh, this one isn't going to auto-merge because the `.mergify.yml` file is modified...one of the guard-rails it has in place.  No worries.

cevich:
Thanks @edsantiago

PR #13, opened 2020-08-28
Only run image builds for PRs

cevich:


lsm5 (dismissed):

cevich:
hrmmm, not sure what's going on here.  The job literally will not run.  Seems like something is broken on the google end.  I may come back to this later.

PR #15, opened 2020-08-28
Fix missing cloud init cfg

cevich:


cevich:
@baude this will fix the disk-partitioning problem in Fedora.

PR #17, opened 2020-09-01
Container images for prior releases

cevich:


cevich:
@baude PTAL.  This adds `prior-fedora_podman` and `prior-ubuntu_podman` container images.  We need these for building podman (and running in_podman testing if need be).

PR #21, opened 2020-09-11
Enable nested-virt for all VM cache-images

cevich:
Ref:
https://cloud.google.com/compute/docs/instances/enable-nested-virtualization-vm-instances

Important Note:  In order to satisfy the >= "Intel Haswell" requirement,
the `enable_nested_virtualization` attribute must be set `true` in the
relevant `gce_instance` block of a Cirrus-CI task.  This is not
documented anywhere I could find, as of this commit.


cevich:
@rhatdan PTAL and submit a Github review.  Approval will lead to this PR to automatically merging.

cevich:
This shouldn't be needed, as non-native VMs don't require CPU support.

PR #23, opened 2020-09-14
Ubuntu download of crun package

cevich:

### DIFF
--- a/cache_images/fedora_packaging.sh
+++ b/cache_images/fedora_packaging.sh
@@ -28,15 +28,6 @@ fi
 
 bigto ooe.sh $SUDO dnf update -y
 
-# Fedora, as of 31, uses cgroups v2 by default. runc does not support
-# cgroups v2, only crun does. (As of 2020-07-30 runc support is
-# forthcoming but not even close to ready yet). To ensure a reliable
-# runtime environment, force-remove runc if it is present.
-# However, because a few other repos. which use these images still need
-# it, ensure the runc package is cached in $PACKAGE_DOWNLOAD_DIR so
-# it may be swap it in when required.
-REMOVE_PACKAGES=(runc)
-
 INSTALL_PACKAGES=(\
     autoconf
     automake
@@ -123,6 +114,7 @@ INSTALL_PACKAGES=(\
     redhat-rpm-config
     rpcbind
     rsync
+    runc
     sed
     skopeo
     skopeo-containers
@@ -152,21 +144,18 @@ else
 fi
 
 
+# Download these package files, but don't install them; Any tests
+# wishing to, may install them using their native tools at runtime.
 DOWNLOAD_PACKAGES=(\
     "cri-o-$(get_kubernetes_version)*"
     cri-tools
     "kubernetes-$(get_kubernetes_version)*"
-    runc
     oci-umount
     parallel
 )
 
 echo "Installing general build/test dependencies"
 bigto ooe.sh $SUDO dnf install -y $EXARG "${INSTALL_PACKAGES[@]}"
-
-if [[ ${#REMOVE_PACKAGES[@]} -gt 0 ]]; then
-    lilto ooe.sh $SUDO dnf erase -y "${REMOVE_PACKAGES[@]}"
-fi
 
 if [[ ${#DOWNLOAD_PACKAGES[@]} -gt 0 ]]; then
     echo "Downloading packages for optional installation at runtime, as needed."
--- a/cache_images/ubuntu_packaging.sh
+++ b/cache_images/ubuntu_packaging.sh
@@ -46,6 +46,7 @@ INSTALL_PACKAGES=(\
     coreutils
     cri-o-runc
     criu
+    crun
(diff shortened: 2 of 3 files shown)


cache_images/ubuntu_packaging.sh
```diff
@@ -118,6 +118,7 @@ INSTALL_PACKAGES=(\
     zlib1g-dev
 )
 DOWNLOAD_PACKAGES=(\
+    crun
```
edsantiago, line 121: Why DOWNLOAD_PACKAGES? Why not in INSTALL_PACKAGES?

cevich: @mheon wanted this so he could test some things.  I didn't want to risk mucking up everybody's tests by having both `runc` and `crun` installed at the same time.  Am I wrong, is that a perfectly safe thing to do?

cevich:
note to me: needs rebase

cache_images/ubuntu_packaging.sh
cevich: more context: None of the Ubuntu VMs have cgv2 enabled by default on the kernel.  @mheon is expecting this.

edsantiago: I have to assume that podman will figure out whether to use runc or crun; but this has to be decided by @mheon.

@cevich, just so @mheon (and I) can understand the full ramifications, can you please ELI5 the difference between DOWNLOAD and INSTALL? I think I have a small understanding of it but don't want to misspeak.

mheon: I don't actually know what the default runtime chosen will be on a cgroupsv1 system with runc and crun both installed, I'd need to look at the code to check...

cevich: Remember too: These VMs are potentially used across multiple projects.  So for example, `buildah` CI would also need to behave with having both installed.  The safest thing is to download it like this, then manage the install/swap at test-time.  This gets consistent/reliable package versions, and won't mess with other automation.  Where were you thinking of testing this from.  Manually or in a script or in ginkgo or somewhere else?

mheon: I have confirmed that `containers.conf` will correctly select `runc` as the default on a cgroups v1 system if both runc and crun are available. Buildah does *not* use this bit of containers.conf (which seems like an unfortunate oversight...) but it, too, correctly uses runc on v1 systems even if both are installed.

I think it ought to be safe to have both runtimes installed given this.

cevich: Thanks guys.  Okay so then for consistency we should probably do the same thing on both Fedora and Ubuntu (install crun & runc).  Once the images are available, I'll open PRs for podman and buildah and some of the other repos.  Both to be on the safe-side, and because many/most projects are using old-ish images.

cache_images/fedora_packaging.sh
```diff
@@ -152,22 +144,19 @@ else
 fi
 
 
+# Packages needed for some special testing cases
+# where the test will perform the installation.
```
edsantiago, line 148: Could I kindly ask for a maintainer-friendly comment here? Something like "we download the rpm for these packages, but don't install them; tests wishing to use them can install them by running dnf install /path/to/where/the/rpm/was/downloaded"

cache_images/ubuntu_packaging.sh
```diff
@@ -117,6 +118,8 @@ INSTALL_PACKAGES=(\
     zip
     zlib1g-dev
 )
+# Packages needed for some special testing cases
+# where the test will perform the installation.
```
edsantiago, line 122: Like the above, but in ubuntu-speak ("apt", etc)

edsantiago (approved):
LGTM, other than two future-friendly comment requests.

I spent many minutes considering whether the `REMOVE_PACKAGES` code should be left in (something I rarely do for dead code!) but no, I agree with your decision to excise it. Let's trust that we never need anything like that again.

cache_images/fedora_packaging.sh
cevich: I surely hope all possible future maintainers know `package==rpm` and use `dnf` to install them (similar for Ubuntu) :rofl:  I've got a ***bunch** of testing to do with these images once they build, so I'd prefer to prioritize that over a comment verbosity fix.  Though if I need to re-roll the images again here for some reason, I'm happy to copy-paste your suggestion in.

edsantiago: What I meant was, it is not clear where the downloads go. But OK, a careful reader can figure it out. Once you get that flake fixed, feel free to merge.

cevich: You're in luck, something else broke.  Adding your suggested comment...

cevich:
Testing new images: b6110627968057344 and c6110627968057344

cevich:
Okay, this seems to be okay (having both runc & crun) in the images.  There were a few hiccups on some repos but they don't appear to be caused by the change.

PR #25, opened 2020-09-17
Add VM-Image management containers

cevich:

### DIFF
--- a/ci/make_container_images.sh
+++ b/ci/make_container_images.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# This script is intended to be used by Cirrus-CI, from the VM
+# built by the 'image_builder' makefile target in this repo.  Use
+# of this script in any other context/environment is unlikely to
+# function as intended.
+
+set -e
+
+SCRIPT_FILEPATH=$(realpath "$0")
+SCRIPT_DIRPATH=$(dirname "$SCRIPT_FILEPATH")
+REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
+
+# shellcheck source=./lib.sh
+source "$REPO_DIRPATH/lib.sh"
+
+# When running under Cirrus-CI on a branch, also tag/push "latest" images
+PUSH_LATEST=0
+
+# shellcheck disable=SC2154
+if [[ -z "$CI" ]] || [[ "$CI" != "true" ]] || [[ "$CIRRUS_CI" != "$CI" ]]; then
+    die "Unexpected \$CI='$CI' and/or \$CIRRUS_CI='$CIRRUS_CI'"
+elif [[ -z "$IMG_SFX" ]]; then
+    die "Script requires non-empty values for \$IMG_SFX='$IMG_SFX'"
+elif [[ -z "$TARGET_NAME" ]]; then
+    die "Script requiresmakefile target \$TARGET_NAME to be non-empty"
+elif [[ -z "$DEST_FQIN" ]]; then
+    die "Script requires destination FQIN \$DEST_FQIN to be non-empty"
+elif [[ $LOGIN_CMD =~ ENCRYPTED ]]; then
+    die "\$LOGIN_CMD did not get decrypted by Cirrus"
+elif [[ ${#LOGIN_CMD} -le 5 ]]; then
+    die "\$LOGIN_CMD is a ${#LOGIN_CMD}-length string; something almost certainly went wrong with Cirrus decryption"
+fi
+
+set -eo pipefail
+make "$TARGET_NAME" IMG_SFX=$IMG_SFX
+
+set +x
+# Cirrus sets CIRRUS_BRANCH="pull/#" for pull-requests, branch-name for branches
+# shellcheck disable=SC2154
+if [[ -n "$CIRRUS_BRANCH" ]] && [[ -z "$CIRRUS_PR" ]] && [[ -z "$CIRRUS_TAG" ]]; then
+    PUSH_LATEST=1
+fi
+
+trap "podman logout --all" EXIT INT CONT
+# Out of pure laziness, the entire command is encrypted in Cirrus-CI
+${LOGIN_CMD}
+
+set -x
+SRC_FQIN="$TARGET_NAME:$IMG_SFX"
+podman tag "$SRC_FQIN" "$DEST_FQIN"
+podman push "$DEST_FQIN"
+
+if ((PUSH_LATEST)); then
+    LATEST_FQIN="${DEST_FQIN%:*}:latest"
+    podman tag "$SRC_FQIN" "$LATEST_FQIN"
+    podman push "$LATEST_FQIN"
+fi
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -57,7 +57,7 @@ task:
 
 
 task:
-    alias: "podman_containers"
+    alias: "container_images"
     depends_on:
         - image_builder
     only_if: $CIRRUS_PR =~ '.+'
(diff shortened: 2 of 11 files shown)


cevich:
@edsantiago when you have a moment PTAL, there's only a few new changes/files, the rest are a nearly identical port from the `containers/podman` repo.  Specifically from `contrib/imgts` and `contrib/imgprune`.  New changes/files here are:

* [`.cirrus.yml`](https://github.com/containers/automation_images/pull/25/files#diff-0f7067c31caedf7df2739c9b07efa122)
* [`Makefile`](https://github.com/containers/automation_images/pull/25/files#diff-b67911656ef5d18c4ae36cb6741b7965)
* [`ci/make_container_images.sh`](https://github.com/containers/automation_images/pull/25/files#diff-fdbc7905a0726730ec720a73d7e5b4ad)
* [one `readme`](https://github.com/containers/automation_images/pull/25/files#diff-36864226d368d83c0172b369e67ef8c9)
* [other `readme`](https://github.com/containers/automation_images/pull/25/files#diff-6f27b20036c6a87b103196a291855b67)

P.S. [I tested the imgts container in the podman PR, where I left imgprune disabled for now while I ponder a safer solution.](https://github.com/containers/podman/pull/7711)

ci/make_container_images.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# shellcheck source=./lib.sh
+source "$REPO_DIRPATH/lib.sh"
+
+# When running under Cirrus-CI on a branch, also tag/push "latest" images
+PUSH_LATEST=0
+
+# shellcheck disable=SC2154
+if [[ -z "$CI" ]] || [[ "$CI" != "true" ]] || [[ "$CIRRUS_CI" != "$CI" ]]; then
+    die "Unexpected \$CI=$CI and/or \$CIRRUS_CI=$CIRRUS_CI"
```
edsantiago, line 22: A good habit when displaying variables is to show them in quotes:
```
   ... \$CI='$CI' and/or ...
```
This way if the variable is empty, or has a space, or an equals, or just generally is weird, the value is clearer to the reader.

ci/make_container_images.sh
```diff
@@ -0,0 +1,56 @@
...
+elif [[ -z "$IMG_SFX" ]]; then
+    die "Required non-empty values for \$IMG_SFX=$IMG_SFX"
```
edsantiago, line 24: Can you make these error messages clearer to the reader? "Required ..." does not give me a good indication of what the problem is. Perhaps simply "Required envariable \$IMG_SFX is undefined". Likewise for the next two.

ci/make_container_images.sh
```diff
@@ -0,0 +1,56 @@
...
+elif [[ -z "$TARGET_NAME" ]]; then
+    die "Required makefile target \$TARGET_NAME to be non-empty"
+elif [[ -z "$DEST_FQIN" ]]; then
+    die "Required destination FQIN \$DEST_FQIN to be non-empty"
+# avoid exposing secret contents
+elif [[ "${#LOGIN_CMD}" -le 0 ]] || grep -Eq "ENCRYPTED" <<<"$LOGIN_CMD"; then
```
edsantiago, line 30: This is really hard to understand: under what circumstances will a string length be less than zero? Why are you using `grep -E` on something that is in no way a regex (and could better be handled by fgrep)? And, this isn't running under `set -x`, so where is the risk of exposure? Couldn't this be handled via:
```bash
elif [[ $LOGIN_CMD =~ ENCRYPTED ]]; then
    die "\$LOGIN_CMD did not get decrypted by Cirrus"
elif [[ ${#LOGIN_CMD} -le 5 ]]; then
    die "\$LOGIN_CMD is a ${#LOGIN_CMD}-length string; something almost certainly went wrong with Cirrus decryption"
fi
```

ci/make_container_images.sh
```diff
@@ -0,0 +1,56 @@
...
+
+set +x
+# Cirrus sets CIRRUS_BRANCH="pull/#" for pull-requests, branch-name for branches
+# shellcheck disable=SC2154
+if [[ -n "$CIRRUS_BRANCH" ]] && [[ -z "$CIRRUS_PR" ]] && [[ -z "$CIRRUS_TAG" ]]; then
+    PUSH_LATEST=1
+fi
+
+trap "podman logout --all" EXIT INT CONT
+${LOGIN_CMD}
```
edsantiago, line 45: This is not something to fix now, but definitely in future such work: I am very uncomfortable with the "podman logout". Up until this line I had no idea what \$LOGIN_CMD was, I just shrugged and let it pass. Now I realize it must be a "podman login" command ... and it just feels wrong to have a "podman logout" without a matching login. In future work I think it would be more maintainable to do:
```bash
    podman login --user $DECRYPTED_PODMAN_REGISTRY_USER \
        --password-stdin \
        $REMOTE_REGISTRY <<<"$DECRYPTED_PODMAN_REGISTRY_PASSWORD"
```
(with better variable names of course; those are just for clarity here)

edsantiago (commented):
I mostly reviewed via `diff -ur ../podman/contrib/{imgts,imgprune}`. Other than notes below, first pass LGTM.

ci/make_container_images.sh
cevich: oh yes, I usually try to do that.  Good catch.

ci/make_container_images.sh
cevich: Oh one missing bit of context here is: the `die()`  function will prefix the message with `ERROR:`.  Does that sway your opinion?

ci/make_container_images.sh
cevich: > so where is the risk of exposure

Myself, I've done this:  Debugging and do a `set -x` on the whole script then run it through CI :confounded: 

Otherwise your suggestion is sound/good idea.

ci/make_container_images.sh
cevich: FWIW: `podman logout` always works, even if you're not logged in.  I put everything in one var. out of pure laziness :smile: Your way is certainly more readable though, I get your point.

cevich:
Thanks for taking a peek @edsantiago I'll make those few fixes and re-run the build just to make sure it passes.  Otherwise I've already got the image-suffix `c6106205175152640` referenced in other PRs.  No matter though, it seems we'll be needing fresh images in podman/buildah anyway to address the `fchmodat2` thing.

ci/make_container_images.sh
edsantiago: No, my concern is that "Required" is ambiguous in this context: it isn't clear if it's an adjective or a past-tense verb. Someone reading that message (like me) might interpret it as "The required non-empty values for \$IMG_SFX ... [then what? incomplete sentence?]". (Also, it doesn't seem very helpful to add `=$VAR` when you've just confirmed that `$VAR` is empty). If you like the word, may I suggest "I require non-empty values for..." or "this tool requires non-empty values for..."?

cevich:
fixed & force-pushed.  @edsantiago if/when you're ready drop an approving review and mergify will merge this.  Thanks again.

ci/make_container_images.sh
```diff
@@ -0,0 +1,58 @@
...
+source "$REPO_DIRPATH/lib.sh"
+
+# When running under Cirrus-CI on a branch, also tag/push "latest" images
+PUSH_LATEST=0
+
+# shellcheck disable=SC2154
+if [[ -z "$CI" ]] || [[ "$CI" != "true" ]] || [[ "$CIRRUS_CI" != "$CI" ]]; then
+    die "Unexpected \$CI='$CI' and/or \$CIRRUS_CI='$CIRRUS_CI'"
+elif [[ -z "$IMG_SFX" ]]; then
+    die "Script requires non-empty values for \$IMG_SFX='$IMG_SFX'"
```
edsantiago, line 24: This is still an unhelpful `=$VAR`

ci/make_container_images.sh
```diff
@@ -0,0 +1,58 @@
...
+elif [[ -z "$TARGET_NAME" ]]; then
+    die "Script requiresmakefile target \$TARGET_NAME to be non-empty"
```
edsantiago, line 26: missingspace

edsantiago (approved):
Nits can be addressed in future. LGTM.

PR #28, opened 2020-09-30
Use fixed-version of BATS

cevich:
Due to bug https://bugs.launchpad.net/ubuntu/+source/bats/+bug/1882542
which also affects setting things like `$IFS` a fixed version of Bats
is installed from source.  This is done using the containers/podman
Makefile, `install.tools` target for ease of maintenance.

See also: https://github.com/containers/podman/pull/6515

### DIFF
--- a/cache_images/fedora_packaging.sh
+++ b/cache_images/fedora_packaging.sh
@@ -32,7 +32,6 @@ INSTALL_PACKAGES=(\
     autoconf
     automake
     bash-completion
-    bats
     bridge-utils
     btrfs-progs-devel
     buildah
--- a/cache_images/ubuntu_packaging.sh
+++ b/cache_images/ubuntu_packaging.sh
@@ -34,7 +34,6 @@ INSTALL_PACKAGES=(\
     autoconf
     automake
     bash-completion
-    bats
     bison
     btrfs-progs
     build-essential


edsantiago:
...and, for those viewers playing along at home wondering why there's no `make install.tools` addition via this PR, or (when finding that it's already run by default) why it didn't override the host-provided bats: the `install.bats` target checks `type -t bats` and only triggers if there is no `bats` executable in $PATH. (I knew this, because I reviewed the PR that did this a few weeks ago, and it still confused me even so).

For future spelunkers: the symptom is that Ubuntu has forked BATS, and their version has a horrible bug (see above) in which environment variables are not preserved _even if defined in a bats-file top-level context_. The symptom you will see is that all `login` tests fail, because each test tries to contact the registry on a different random port, because the global `$PODMAN_LOGIN_REGISTRY_PORT` is not being propagated.

Anyhow, LGTM.

edsantiago (approved):
Oh wait this isn't a /lgtm repo

cevich:
@edsantiago thanks for the helpful commentary, may it help to alleviate any possible future confusions.

cevich:
Testing in https://github.com/containers/podman/pull/6784

PR #29, opened 2020-10-12
Re-implement image life-cycle management

cevich:
Previously, successfully updating VM image metadata was treated as
optional (errors ignored).  This resulted in unrelated failures
(e.g. auth. token expiring) going unnoticed.  Consequentially VM
images that were in-use got deleted without warning.  Fix this by
making the operations within `imgts` mandatory (do not ignore
failures).

To further protect VM images from being removed while in use, introduce
a new container image `imgobsolete`, to be run before `imgprune`.
Instead of deleting images, they will be marked as 'deprecated' and
'obsolete' by this new container.  It will also encode advisory
deletion-date metadata, set 30 days into the future.

Lastly, update the `imgprune` container image such that it only searches for
images marked 'deprecated' and 'obsolete' with a metadata deletion-date
in the past.  This greatly simplifies the script.

The new 'obsolete' image state prevents the image from being used
or listed, but does not actually remove them.  If a mistake is
noticed, the image can be recovered manually with a simple
`gcloud` command:

```
gcloud compute images deprecate <NAME> --state=ACTIVE
```

### DIFF
--- a/imgprune/entrypoint.sh
+++ b/imgprune/entrypoint.sh
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+# This script is set as, and intended to run as the `imgprune` container's
+# entrypoint.  It is largely based on the imgobsolete's entrypoint script
+# but with some important/subtle differences.  It searches for deprecated
+# VM images with deletion-metadata some time in the past.  Some number of
+# these images are randomly selected and then permanently deleted.
+
 set -e
 
 # shellcheck source=imgts/lib_entrypoint.sh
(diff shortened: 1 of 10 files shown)


cevich:
@edsantiago when you have a moment PTAL...This PR is a step toward re-enabling safer VM image cleanup, with the addition of an "undelete" option.  I apologize in advance, as it's quite "heady", with some non-linear flows.  I've done my best to dot i's, cross t's, and comment the heck out of every tricky spot.  

Your pretty-graph drawing script works, but must be understood as running under three different modes:
dependencies

1. (Top most flow) Running as part of a PR, after images are built: Apply metadata to fresh VM Images to verify `imgts` is functioning.  Followup with a dry-run of `imgobsolete` and `imgprune` (few other practical way to test these).

2. (Bottom then middle flow) Running on a branch after push or PR merge:  Build "latest" tagged versions of all container images only, no VM images are built.  No tests run (but they could).

3. (Bottom-only flow) Running on a branch triggered by cron: No images are built.  Use the "latest" tagged images for `imgobsolete` followed by `imgprune`.

Hopefully that makes sense.

cevich:
Idea: There's no way to test the branch-push and cron workflows.  If either/both break, I'll just have to deal with it in followup PRs.

imgprune/entrypoint.sh
```diff
@@ -9,85 +15,52 @@ req_env_var GCPJSON GCPNAME GCPPROJECT
...
+# Field list from `gcloud compute images list --limit=1 --format=text`
+FORMAT='value[quote](name,selfLink,deprecated.state,deprecated.deleted,labels)'
 # Required variable set by caller
 # shellcheck disable=SC2154
 PROJRE="/v1/projects/$GCPPROJECT/global/"
-RECENTLY=$(date --date='3 days ago' --iso-8601=date)
 # Filter Ref: https://cloud.google.com/sdk/gcloud/reference/topic/filters
-# shellcheck disable=SC2154
-FILTER="selfLink~$PROJRE AND creationTimestamp<$RECENTLY"
+FILTER="selfLink~$PROJRE AND deprecated.state=OBSOLETE AND deprecated.deleted>$ABOUTNOW"
```
edsantiago, line 30: I'm finding it impossible to find examples of `deprecated.deleted` (or `deprecated.anything`), and I've spent far too long on this small piece already, but should that `>` (greater-than) be `<` (less-than)? If it's correct as it is, could you maybe add a comment explaining what `deprecated.deleted` means? And maybe add a corresponding comment in `imgobsolete/entrypoint.sh#L101` if it somehow corresponds to the `--delete-in=30d` therein?

edsantiago:
Whew. As best I can tell, with the exception of my `<` vs `>` question above, it looks like it should work. I have a bunch of changes in my local repo, where I've edited some comments and variables to help me better understand what's going on; I may clean that up and submit it next week, because I think it might help a future maintainer who is less well versed in gcloud-fu than you are.

BTW the changes I made last week to cirrus-task-map, for displaying SKIP/ONLY, needed only a few seconds' work to produce a more readable plot that doesn't require a text legend: 
cirrus-map-pr29

edsantiago (approved):
Tentative approval (modulo the `deprecated.deleted` question below). I will not have a chance to look at this again until next week.

edsantiago:
Was **NOT** expecting my "approve" to auto-merge. Damn.

cevich:
> changes I made last week to cirrus-task-map

Suggestion: Can you stick the only_if/skip label on the arrow instead of on the boxes?

> Was NOT expecting my "approve" to auto-merge

It's not a problem, there are as-of-yet no cron-jobs in place that would execute these containers.  Incremental updates to the repo. next week are fine.

cevich:
> Was NOT expecting my "approve" to auto-merge

https://github.com/containers/automation_images/pull/33

edsantiago:
>Suggestion: Can you stick the only_if/skip label on the arrow instead of on the boxes?

That sounded counterintuitive to me, but I gave it a try. End result: no, it's much too confusing. First, in podman, there's an `automation` task that has no precursors but has a skip. Second, also in podman, there are flows that have different skips: like, `unit_test` is skipped on CI:DOC, but `*_integration_test` are skipped both on CI:DOC and on BRANCH. With skips on edges (arrows), my mind wants to think "oh, once I see a skip here, I don't have to look farther right", and that intuition is wrong. Finally, and most importantly to me, it is the task that is skipped, not the flow. It was worth a quick try, but no, I don't think it works well. Thanks for the suggestion.

cevich:
> Thanks for the suggestion.

Thanks for trying it out.

PR #33, opened 2020-10-22
[CI:DOCS] Add pull-request template

cevich:
This is intended to draw attention to the automatic-merge configuration

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -38,6 +38,7 @@ validate_task:
 image_builder_task:
     name: "Image-builder image"
     alias: "image_builder"
+    only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
     depends_on:
         - validate
     # Packer needs time to clean up partially created VM images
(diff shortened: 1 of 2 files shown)


cevich:
@edsantiago PTAL when you get a chance.  This should help POLA re: automatic merging of PRs.

Ref: https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository

edsantiago:
I can't approve this: it basically forces the user to memorize and use an arcane special magic process on each PR. Is there some way we can use the slash-lgtm convention to be consistent across containers repos?

cevich:
> Is there some way we can use the slash-lgtm convention to be consistent across containers repos?

That's far to big of a hammer or such a small repo. touched by 3-4 people (if that).  For the same reason, it doesn't make sense to require >1 approval.  So in the end, the ultimate result of both mergify and prow+tide (`/lgtm`) is exactly the same: The PR merges.

I was hoping this PR would be more of a middle-ground: Sure there's some "astonishment" (compared to a few other containers repos.), but at least it's documented in PR author's faces (beyond just the `README.md`)

I considered disabling mergify here, and using manual merges, but I think the benefit are marginal.  A few more seconds of thinking before manually pressing a button?  Frankly it's just easier for me to simply remind reviewers every time.

PR #31, opened 2020-10-22
WIP: Deprecate prior-ubuntu image builds

cevich:
Alternative to https://github.com/containers/automation_images/pull/34, simply drop support for automation on multiple Ubuntu releases.  Stick to only using the most recent.  This change will require ALL automation, in ALL repositories using these Ubuntu images to drop "prior-ubutnu" references.  e.g. `$PRIOR_UBUNTU_CACHE_IMAGE_NAME` in podman's CI.

PR #34, opened 2020-10-23
Use complete Ubuntu release + update base images

cevich:
As an alternative to https://github.com/containers/automation_images/pull/31, support multiple Ubuntu `v20.X` versions.  This change will require ***ALL*** automation, in ***ALL*** repositories using these Ubuntu images to mirror the definition change:

```diff
diff --git a/lib.sh b/lib.sh
index 58f28fe..cf9ada3 100644
--- a/lib.sh
+++ b/lib.sh
@@ -11,7 +11,7 @@ SCRIPT_DIRPATH=$(dirname "$SCRIPT_FILEPATH")
 # By default, assume we're not running inside a container
 CONTAINER="${CONTAINER:-0}"

-OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | cut -d '.' -f 1)"
+OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | tr -d '.')"
 OS_RELEASE_ID="$(source /etc/os-release; echo $ID)"
 OS_REL_VER="$OS_RELEASE_ID-$OS_RELEASE_VER"

```

Also, ***ANY/ALL*** relevant references to `$OS_RELEASE_VER` values must be updated, for example `19` -> `1910`, `20` -> `2004` or `2010`.


cevich:
Note to me: OBS currently missing Ubuntu 20.10 packages

lsm5:
> Note to me: OBS currently missing Ubuntu 20.10 packages

intentional for now, unless we get more contributors/volunteers.

cevich:
Note to me: Need criu packages for Ubuntu 20.10

lsm5:
> Note to me: Need criu packages for Ubuntu 20.10

I never packaged criu for OBS. I think it would be in ubuntu repos by default, no?

cevich:
It might be for 20.10, previously it came from a Launchpad repo.

cevich:
@lsm5 confirmed, `criu` is now available in 20.10, no need to use launchpad.

cevich:
@lsm5 I'm hitting a major problem with Ubuntu 20.10 packages (we decided not to use OBS...if you remember).  Specifically the version of runc is older then the dinosaurs:

runc-1.0.0~rc10-0ubuntu2-amd64

For reference, even the old F32 VMs have a newer version:

runc-1.0.0-144.dev.gite6555cc.fc32-x86_64

Who/How/Where do I need to deposit the requisite number of cookies in order for this to be updated?

cevich:
@lsm5 un-ping.  I'm going to try enabling cgroupsv2 in Ubuntu 20.10...

cevich:
note to me: waiting for fix to https://bodhi.fedoraproject.org/updates/FEDORA-2020-f5e1592682

cevich:
c5325336827133952

cache_images/ubuntu_packaging.sh
```diff
@@ -20,12 +20,24 @@ lilto ooe.sh $SUDO apt-get -qq -y update
...
+
+# The versions of some key dependencies are not updated quickly
+# enough to support upstream containers-org development.
 VERSION_ID=$(source /etc/os-release; echo $VERSION_ID)
-echo "deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_$VERSION_ID/ /" \
-    | ooe.sh $SUDO tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
-ooe.sh curl -L -o /tmp/Release.key "https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_${VERSION_ID}/Release.key"
-ooe.sh $SUDO apt-key add - < /tmp/Release.key
+REPO_URL="https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/testing:/ci/xUbuntu_$VERSION_ID/"
+GPG_URL="https://build.opensuse.org/projects/devel:kubic:libcontainers:testing:ci/public_key"
```
lsm5, line 35: Are you sure it's `public_key`? If you look at https://software.opensuse.org//download.html?project=devel%3Akubic%3Alibcontainers%3Atesting%3Aci&package=runc , it still says `Release.key` and not `public_key`.

cache_images/ubuntu_packaging.sh
```diff
@@ -79,13 +89,12 @@ INSTALL_PACKAGES=(\
     libnl-3-dev
     libprotobuf-c-dev
     libprotobuf-dev
+    libseccomp
     libseccomp-dev
-    libseccomp2
```
lsm5, line 83: Package name should still be `libseccomp2` and not `libseccomp`. See https://gitlab.com/rhcontainerbot/libseccomp/-/blob/debian/debian/control . `libseccomp` is just the main/source package. The generated binary/library is still libseccomp2.

lsm5 (dismissed):

cevich:
`c6233039174893568`

cevich:
***Note***: These images are passing tests in podman now.  Giving this a positive github review will cause this PR to automatically merge.

edsantiago (changes requested):
Some suggestions for making this a little easier to understand and maintain

cache_images/fedora_setup.sh
```diff
@@ -15,23 +15,34 @@ REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
...
 # (packer defines this envar)
 # shellcheck disable=SC2154
-if [[ "$PACKER_BUILD_NAME" =~ prior ]]; then
-    ENABLE_UPDATES_TESTING=0
-else
-    ENABLE_UPDATES_TESTING=1
-fi
+#if [[ "$PACKER_BUILD_NAME" =~ prior ]]; then
+#    ENABLE_UPDATES_TESTING=0
+#fi
```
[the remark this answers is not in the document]
cevich: Actually, the package may have made it out of updates-testing by now, perhaps I can just un-comment this.

cache_images/fedora_setup.sh
```diff
@@ -15,23 +15,34 @@ REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
...
+
 custom_cloud_init
 
+
+if ! ((CONTAINER)) && [[ "$OS_RELEASE_VER" -le 32 ]]; then
+    warn "Disabling CgroupsV2 kernel command-line option for systemd"
+    SEDCMD='s/^GRUB_CMDLINE_LINUX="(.*)"/GRUB_CMDLINE_LINUX="\1 systemd.unified_cgroup_hierarchy=0"/'
+    ooe.sh $SUDO sed -re "$SEDCMD" -i /etc/default/grub
+    # This is always a symlink to the correct location under /boot/...
+    ooe.sh $SUDO grub2-mkconfig -o $(sudo realpath --physical /etc/grub2.cfg)
```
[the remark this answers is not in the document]
cevich: oh good catch. thanks.

cache_images/ubuntu_packaging.sh
```diff
@@ -20,12 +20,25 @@ lilto ooe.sh $SUDO apt-get -qq -y update
 bigto ooe.sh $SUDO apt-get -qq -y upgrade
 
 echo "Configuring additional package repositories"
-lilto ooe.sh $SUDO add-apt-repository --yes ppa:criu/ppa
+
+
+if [[ "$OS_RELEASE_VER" -le 2004 ]]; then
```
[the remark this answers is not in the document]
cevich: Oh, the comment is right below that, but I can see how it would be confusing.  I'll update it to be clearer.

cache_images/ubuntu_setup.sh
```diff
@@ -9,17 +9,24 @@ SCRIPT_FILEPATH=$(realpath "$0")
...
 source "$REPO_DIRPATH/lib.sh"
 
 bash $SCRIPT_DIRPATH/ubuntu_packaging.sh
 
 if ! ((CONTAINER)); then
-    echo "Making Ubuntu kernel to enable cgroup swap accounting as it is not the default."
-    SEDCMD='s/^GRUB_CMDLINE_LINUX="(.*)"/GRUB_CMDLINE_LINUX="\1 cgroup_enable=memory swapaccount=1"/g'
+    warn "Making Ubuntu kernel to enable cgroup swap accounting"
+    SEDCMD='s/^GRUB_CMDLINE_LINUX="(.*)"/GRUB_CMDLINE_LINUX="\1 cgroup_enable=memory swapaccount=1'
+    if [[ "$OS_RELEASE_VER" -ge 2010 ]]; then
```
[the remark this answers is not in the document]
cevich: Yes, but tricky to fix as not all build context have easy access to a value containing "prior" in it.  This will require some craftiness to fix...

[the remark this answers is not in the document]
edsantiago: Can you get it from `$VM_IMAGE_NAME`? 

(Hint: view-source on any logformatter-processed result, e.g. [this](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6188572304736256/html/int-podman-ubuntu-2004-root-host.log.html), includes a data dump of envariables near the top).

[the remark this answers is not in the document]
cevich: okay, it's convoluted, but I think I found a way to make checking a match against "prior" workable.

cache_images/ubuntu_packaging.sh
cevich: fixed

cache_images/ubuntu_packaging.sh
cevich: fixed

cevich:
***Note:*** Marking as a draft to block automatic merging. I'd like to confirm (at least in a test PR) that the new changes don't break podman testing.

Makefile
```diff
@@ -220,6 +219,7 @@ $(_TEMPDIR)/%_podman.tar: podman/Containerfile podman/setup.sh $(wildcard base_i
 	podman build -t $*_podman:$(call err_if_empty,IMG_SFX) \
 		--build-arg=BASE_NAME=$(subst prior-,,$*) \
 		--build-arg=BASE_TAG=$(call err_if_empty,BASE_TAG) \
+		--build-arg=PACKER_BUILD_NAME=$(subst _podman,,$*) \
```
cevich, line 222: Re: Scripts using conditionals based on presence/absence of 'prior' appearing in the VM/Container image name.

> Can you get it from $VM_IMAGE_NAME?

That var. doesn't exist here (completely different Cirrus-CI setup).  In any case, it's value is derived from `$PACKER_BUILD_NAME` when producing VM images.  The problem area is with the container build workflow, as it shares the same setup/packaging scripts as the VM setup but does ***not*** use the packer tool.

Ordinarily the scripts could ignore this, as all conditions on `$PACKER_BUILD_NAME` only applied to VM's (i.e. containers don't care about setting kernel options)...except now this isn't the case, since I've un-commented the (previously commented in this PR) `$ENABLE_UPDATES_TESTING` conditional (on 'prior' flavors) for fedora.

I don't like this fix, it's convoluted and non-obvious :astonished: (esp. in the container image build case).  However, right now I cannot think of a better way to solve this in a more uniform yet readable manner, and am loathe to add a giant comment to the Makefile (which s already very dense and complex) :disappointed: 

@edsantiago  I'd really like to get this PR merged, since the previously built images were already rolled out all across container-org. CI-land.  Is this something we should just live with, fix later, ignore, or maybe I could file an issue on it (here) for fixing next year?

cevich: Ooof...and it doesn't appear to work: [F32 container build log](https://cirrus-ci.com/task/6207579749613568?command=main#L598) *sigh*

cevich: @edsantiago nvm my question, this is broken anyway so I'll clean it up and comment the heck out of it.

cevich:
> I thought it would be easy. Do what you need to do; revert back to the hardcoded 20-etc, that's fine.

Except now that I'm aware of the problem and wrote a good comment...it would keep me up a night worrying someone might uncomment/un-hard-code this later and not realize the implications :disappointed: 

I need to re-roll and re-test the images here for the other fixes anyway, so, it's just one more thing (and I confirmed by local testing that the `$PACKER_BUILD_NAME` check works).

cevich:
`c5026195240648704`

cevich:
[Okay, everything seems to pass with `c5026195240648704`](https://github.com/containers/podman/pull/8765/checks)

PR #37, opened 2020-11-05
Only build images in PRs

cevich:
There's no useful workflow based on images built during branch or tag
push, only pull-requests.  Further, there is no reason to build anything
during Cirrus-Ci cron-triggered builds.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,6 +23,7 @@ timeout_in: 60m
 validate_task:
     name: "Validate"
     alias: "validate"
+    only_if: &is_pr $CIRRUS_PR != ''
     timeout_in: 5m
     container: &ci_container
         # Ref: https://cirrus-ci.org/guide/docker-builder-vm/#dockerfile-as-a-ci-environment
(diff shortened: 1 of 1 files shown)


cevich:
@edsantiago PTAL when u have a second, this is just a minor fix. Approval will merge (since tests all passed).  This is to speed up the cron-based VM Image lifecycle management jobs.  There's no reason for them to be building any images, they simply reuse the 'latest' tagged images produced after the most recent successful run (in a PR).

edsantiago:
So IIUC no tasks will run on merge? Basically, only PR and cron Is that what you expect? If so, would you consider removing `depends_on: container_images` from `cron_imgobsolete_task`? That would make what I think is a more readable graph:
Current: cirrus-map-pr37
Better: 
cirrus-map-pr37-better

cevich:
> So IIUC no tasks will run on merge? Basically, only PR and cron Is that what you expect?

Correct...despite it leaving a somewhat odd taste in my mouth.  Yes, I'm happy to change the dependency tree and agree it will make things clearer.  Thanks for the suggestion..

cevich:
(Flipped draft-status to invalidate "pending review" status.)

cevich:
Reminder: An "approving" review will cause this to merge, since the tests all passed.

edsantiago (approved):
LGTM

PR #38, opened 2020-11-10
Lock to static automation library version

cevich:
This permits breaking changes to merge into `containers/automation`
without unexpectidly breaking any downstream usage.

Ref: https://github.com/containers/automation/pull/58

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -24,8 +24,7 @@ CUSTOM_CLOUD_CONFIG_DEFAULTS="$SCRIPT_DIRPATH/cloud-init/$OS_RELEASE_ID/cloud.cf
 # This location is checked by automation in other repos, please do not change.
 PACKAGE_DOWNLOAD_DIR=/var/cache/download
 
-# TODO: Lock down to specific version number for stability
-INSTALL_AUTOMATION_VERSION="latest"
+INSTALL_AUTOMATION_VERSION="1.2.3"
 
 SUDO="env"
 if [[ "$UID" -ne 0 ]]; then


cevich:
@edsantiago when you get a sec.  Mind popping an 'approved' here so it can merge?

edsantiago:
@cevich how urgent is it? I took a look earlier and this one-line change is much too complicated for me to understand, because I traced it to a `curl` which fetches a huge script and passes the version as an argument which I have no idea how it's used. It's going to take me a long time to understand all that. I don't even know where the `1.2.3` is defined.

If you need this urgently, I can hit approve, but it will be a blind approval without even the slightest review.

cevich:
> how urgent is it?

Not urgent and no need for deep review.  The automation installer script and library was designed to be used this way from the beginning.  We use to specify a fixed version, but I did away with that for podman's CI 2.0 in case rapid changes were needed.  Now, going forward, IMHO stability of the interface is more important - and also because I want to make some [needed, but interface breaking changes.](https://github.com/containers/automation/pull/58)

lib.sh
```diff
@@ -24,8 +24,7 @@ CUSTOM_CLOUD_CONFIG_DEFAULTS="$SCRIPT_DIRPATH/cloud-init/$OS_RELEASE_ID/cloud.cf
 # This location is checked by automation in other repos, please do not change.
 PACKAGE_DOWNLOAD_DIR=/var/cache/download
 
-# TODO: Lock down to specific version number for stability
-INSTALL_AUTOMATION_VERSION="latest"
```
edsantiago, line 28: How did this ever work? As I read the `curl` script, it fails if invoked with "latest"?
https://github.com/containers/automation/blob/ed872b6427d727adf58635d3b3ae23def2bda1d0/bin/install_automation.sh#L91

(It's been a long day, I'm probably just missing something obvious because I don't know how this script is supposed to be invoked)

edsantiago (approved):
Whew. I've read enough of `install_automation.sh` to realize that reviewing it would take me 2-3 days. So instead I've just verified that there really is a `1.2.3` tag in the `containers-automation` repo, and will trust that the curl script will just do the right thing.

cevich:
> reviewing it would take me 2-3 days

Thanks for looking, and I'm glad you realized the scope.   It's def. not worth it, I'm perfectly comfortable relying on the installer being fairly well unit-tested.

> and will trust that the curl script will just do the right thing.

The script is executed as part of image building here, so having the builds not fail can also be taken as reassurance.

Thanks @edsantiago

PR #41, opened 2020-12-08
Fix label update clash

cevich:
According to google-cloud support, over the past year, less than 1%
of VM Image label updates clashed with other simultaneous label updates.
When this occurs, an error and message is returned by the gcloud
command:

```
ERROR: (gcloud.compute.images.update) HTTPError 412: Labels fingerprint
either invalid or resource labels have changed
```

Perhaps worse, this causes the label update job in automation to fail,
and require attention from the developer.  Avoid this situation by
detecting this specific condition, issuing a warning, then ignoring it.

This is safe because:

* There are no other automation processes or manual provisions currently
  in use which touch VM Image labels in any way.
* The label update operations are atomic on the cloud-side, this is how
  they are able to detect the error occurring.
* Image lifecycle management automation operates at 1-day granularity.
  Being a few seconds, minutes, or even hours off is inconsequential.

Therefore it is a reasonably safe assumption that another timestamp
update operations caused the clash and the error can simply be ignored.

### DIFF
--- a/imgts/entrypoint.sh
+++ b/imgts/entrypoint.sh
@@ -31,13 +31,25 @@ ARGS="
 # Don't allow one bad apple to ruin the whole batch
 ERRIMGS=''
 
+# It's possible for multiple simultaneous label updates to clash
+CLASHMSG='Labels fingerprint either invalid or resource labels have changed'
+
 # Must be defined by the cirrus-ci job using the container
 # shellcheck disable=SC2154
 for image in $IMGNAMES
 do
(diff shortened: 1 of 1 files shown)


cevich:
@edsantiago this is not critical, but PTAL when you have a moment.  I think my bash updates will work, but would appreciate the oversight.  Thanks.

cevich:
(note: I canceled the testing, so even if you approve, mergify will not automatically merge this.)

edsantiago (approved):
LGTM. Thanks for tracking this one down.

PR #43, opened 2021-01-07
Add container image for uploading files to GCS

cevich:
Also update other container image directories to correct example build
instructions.


cevich:
note to me: add a "test" cirrus job that uploads a file to verify functionality

PR #56, opened 2021-03-22
Add `zstd` to images

nalind:
We've got tests that need to use the `zstd` CLI tool coming soon.


cevich:
LGTM

The newly built images have id c6374915067609088 for reference and are ready-to-go.

PR #60, opened 2021-04-06
Add upstream golang for ubuntu

baude:
Remove ubuntu golang and add upstream golang to avoid their
non-performant golang

Signed-off-by: baude <bbaude@redhat.com>


cevich (approved):
LGTM - go ahead and merge this if it passes muster in podman CI land.

PR #69, opened 2021-05-03
EXPERIMENTAL: Switch to using kickstart for Fedora base images

cevich:
This is required to change the root filesystem type.  The current
(F33/F34) default for cloud images is ext4, but XFS is desired for
testing.


cevich:
Closing this for now, since there's a LOT of additional complexity in the workflow for apparently minimal performance gain. Happy to revisit/re-open this at a later date if there is a need.

PR #79, opened 2021-06-17
[CI:TOOLING] orphanvms: trigger on start instead of create time

cevich:
There are cases where users have "play" VMs that are normally shutdown
when not in use.  The monitor was falsely triggering on them when
running, due to their very long create-time based age.  Switch to using
the start-time, which will be more accurate in all use cases.


cevich:
I'm just going to merge this, I already tested it manually.

PR #81, opened 2021-06-24
Enable future use of nested VMs

cevich:
This may be necessary for example, for testing out NUMA features.
Google doesn't provide any control over NUMA for VMs it provisions.


cevich:
Okay it works.  I'm just going to merge this and forget about it.  There's zero extra cost to enabling this, and perhaps one day it will be needed.

PR #82, opened 2021-07-12
Commit to trigger image build

cevich:


cevich:
Going with https://github.com/containers/automation_images/pull/84 (which contains an actual/needed change)

PR #87, opened 2021-08-26
Fix workaround due to obsolete update

cevich:


cevich:
Note: I didn't deploy this image anywhere.  This is just in case we need a newer image before the final buildah package update actually hits the repos in a few more weeks.

PR #88, opened 2021-09-08
Add workaround for nm/cni flake + Drop all prior-ubuntu builds

cevich:
This adds a simple (but hopefully effective) workaround for a flake observed in CI across multiple repositories:  NetworkManger tries to control CNI manged interfaces, wrecking havoc on tests when the stars align properly.

Also, it was decided by the team to drop use of 'prior-ubuntu' in CI due to EOL-before-next-release conflicts and needing to maintain/use a custom set of "newer" dependencies for the old releases.  In other words, there's always a perpetually growing divide between 'as-tested' and 'as-used' for the older Ubuntu releases and this will never go away.  Since the Fedora policy is to always support the last-two releases, this will never be an issue there.

Both of these changes are being included in one PR because it makes it easier to incorporate them across all of CI.


cevich:
Note to me: This should go into podman, buildah, skopeo, then also everything else (assuming the first three pass).

cevich:
Note: Need to update images on all repos. anyway, might as well also include decision from https://github.com/containers/podman/issues/11070

edsantiago:
I'd prefer to see the drop-ubuntu commit as a separate PR; was that your intention?

cevich:
> as a separate PR; was that your intention?

Thanks for the feedback, and normally I would agree.  In this case, the "real-work" is getting the new images incorporated into all the `.cirrus.yml` files across all the  repos (and passing all tests).  If one or the other commit needs to be reverted...that's going to involve a new PR in this repo. anyway.  Lastly, not having any 'prior-ubuntu' images with this ID will make it easier to spot missed downstream requirements on it (Cirrus/google will complain loudly about not being able to find the prior-ubuntu image).

Make sense or still concerned?

cevich:
Clarification:  The 'nm/cni' issue is potentially affecting CI everywhere, not just podman.

edsantiago:
My concern was pay-now vs pay-later. The prior-ubuntu thing will need to be reverted one day, to add it back, and it will be harder to find since this PR makes no mention of it whatsoever (except for here: THIS PR REMOVES PRIOR_UBUNTU, but it's not in the title so it's not helpful). But I'm not the one who will be doing that revert, so if you're OK with paying the price then, I'm OK with merging.

/lgtm
/hold

cevich:
> The prior-ubuntu thing will need to be reverted one day

I hope not...Only having one Ubuntu means that many fewer chances for flakes :grin: Seriously though, assuming they stick with the same "4-and-10" release schedule (+ every third is an LTS), it will never line up nicely for us.  There will always eventually be a  4-month long period where one release is EOL - and I'm sick of dealing with that given the minor/marginal benefit it provides downstream to users (due to the needing fully-custom dependency builds).

> this PR makes no mention of it whatsoever

Oh right, good point...and easily fixed :grinning:

TomSweeneyRedHat:
LGTM

cevich:
hrmmm, Cirrus-CI seems to have lost it's marbles :(  Let me try re-pushing.

PR #90, opened 2021-09-15
[CI:TOOLING] Fix skopeo image locale env. vars.

cevich:

### DIFF
--- a/skopeo_cidev/Containerfile
+++ b/skopeo_cidev/Containerfile
@@ -3,9 +3,7 @@ ARG BASE_TAG=latest
 FROM ${BASE_NAME}:${BASE_TAG}
 
 # Required to keep perl & other tooling happy
-ENV LANGUAGE="en_US" \
-    LC_ALL="en_US" \
-    LANG="en_US.UTF-8"
+ENV LC_ALL="C"
 
 COPY /packages.txt /root/
 RUN dnf -y update && \


cevich:
***Note:*** I manually re-tagged and pushed the image as `quay.io/libpod/skopeo_cidev:c6431352024203264` to fix https://github.com/containers/skopeo/pull/1444#issuecomment-919977202

skopeo_cidev/Containerfile
```diff
@@ -3,9 +3,7 @@ ARG BASE_TAG=latest
 FROM ${BASE_NAME}:${BASE_TAG}
 
 # Required to keep perl & other tooling happy
-ENV LANGUAGE="en_US" \
-    LC_ALL="en_US" \
-    LANG="en_US.UTF-8"
+ENV LC_ALL="C"
```
mtrmac, line 6: This might break with non-ASCII characters. Values to consider are `en_US.UTF-8` or perhaps `C.UTF-8`.

(I’m not following this closely, IIRC some of the locale data is not available in default installs, or default containers. I’d expect one of the two above (but _not necessarily `en_US`) to be available. Alternatively, the other locales might be available in packages named something like `glibc-langpack*`. Random googling finds https://github.com/containers/toolbox/issues/60 and https://fedoraproject.org/wiki/Changes/Glibc_locale_subpackaging .)

cevich: Oh, okay.  Installing additional packages isn't a problem.  Whatever's needed so the tests run smoothly.

mtrmac: If `C` works fine, it’s not worth the effort/time to change it yet again for theoretical purity.

I vaguely remember Python in particular being not very happy with `C` + `UTF-8`, but that’s probably not relevant to us.

cevich: Ya, this image is dedicated for skopeo.  Anyway, yes I agree, if it's not a problem, no need to fix it.  Thanks for your help/insight.

PR #96, opened 2021-10-01
Reduce image size slightly

cevich:
Especially for container images, but also for VMs, it's desirable to
not leave a lot of useless junk around.  This is because the images
are compressed and moved around, both of which benefit from a reduced
quantity of bits.  Attempt to improve the situation by selectively
removing/cleaning cached packages and metadata when/where possible.


cevich:
Note to me: *confirm w/n image size is actually smaller*

cevich:
Included into #93

PR #97, opened 2021-10-04
Build push

cevich:
These changes produce a new special-purpose VM image containing an embedded script (`main.sh`) designed to make use of the (included) `build-push.sh` script from c/automation.  The intent is for the skopeo, buildah, and podman repositories to have a shared, uniform, and standard way for building then pushing their own multi-arch manifest lists for public consumption.  To that end, a system-test is included here to confirm expected output, given mock inputs.

***Note:*** Currently this task is performed using [a difficult to read/maintain Github-actions workflow](https://github.com/containers/podman/blob/a08e4e5b4657469a5d686ec5621f681cd240f6e7/.github/workflows/multi-arch-build.yaml), making use of docker and buildx.  Worse, due to design/implementation defects, this workflow is duplicated across all three containers-org repositories.


edsantiago:
Thanks. Yeah, on first pass this is more than I can review right now. I'm on PTO today and getting ready to head out for the rest of the day, will aim to look at it Thursday morning.

cevich:
Thanks, yeah no rush, it's aimed at replacing something already "working", so low-urgency.  I believe you already saw/reviewed the underlying `build-push.sh` script (from c/automation repo).  This PR is just about bundling that tool with an environment and execution recipe.  As-is, it should produce the same output images as the github-action workflows...though I haven't gone as far as actually trying that yet.

build-push/bin/main.sh
```diff
@@ -0,0 +1,133 @@
...
+if ((DRYRUN)); then
+    _DRNOPUSH="--nopush"
+    warn "Operating in dry-run mode with $_DRNOPUSH"
+fi
+
+### MAIN
+
+cd "$BPTMP"
+dbg "Cloning $REPO_URL"
+git clone --no-local --no-hardlinks --depth 1 --single-branch --no-tags "$REPO_URL"
```
cevich, line 69: Note to me: This script is most likely going to be run from CI with $PWD already in the subject repository.  Instead of cloning (again), simply verify `$PWD/.git` exists.  If it doesn't, then do the "clone + cd" operation.

lib.sh
```diff
@@ -37,7 +37,7 @@ if [[ "$OS_RELEASE_ID" == "ubuntu" ]]; then
     SUDO="$SUDO env DEBIAN_FRONTEND=$DEBIAN_FRONTEND"
 fi
 
-if [[ -d "/usr/share/automation" ]]; then
+if [[ -r "/etc/automation_environment" ]]; then
```
edsantiago, line 40: I don't know if this is important/relevant, but there's still one instance of `/usr/share/automation` in the repo:

https://github.com//containers/automation_images/blob/752b774bb3e29fb11e43890f07fd978d8d1eac1e/lib.sh#L63

build-push/bin/main.sh
```diff
@@ -0,0 +1,133 @@
...
+    REPO_URL=$REPO_URL
+    REPO_NAME=$REPO_NAME
+    CTX_SUB=$CTX_SUB
+    CTX_NAME=$CTX_NAME
+    REPO_FQIN=$REPO_FQIN
+"
+
+# Set non-zero to avoid actually executing build-push, simply print
+# the command-line that would have been executed
+DRYRUN=${DRYRUN:0}
```
edsantiago, line 58: Did you mean `:-0` (missing minus sign)?

build-push/bin/main.sh
```diff
@@ -0,0 +1,133 @@
...
+    dbg "parsed version: $img_cmd_version"
+    test -n "$img_cmd_version"
+    lblargs="$lblargs --label=org.opencontainers.image.version=$img_cmd_version"
+    # Prevent temporary build colliding with multi-arch manifest list (built next)
+    # but preserve image (by ID) for use as cache.
+    dbg "Un-tagging $REPO_FQIN"
+    podman untag $REPO_FQIN
+
+    # Stable images get pushed to 'containers' namespace as latest & version-tagged
+    # Common library defines SCRIPT_PATH
```
edsantiago, line 114: I don't understand this comment. Did you mean to apply this to line 47 instead of this one?

lib.sh
cevich: Ya this is fine because that function specifies `INSTALL_PREFIX=/usr/share`.  I only changed it here because `INSTALL_PREFIX` could change, but `/etc/automation_environment` will always be a constant indicator that the library/tooling has been installed.

build-push/bin/main.sh
cevich: Oh, yes I did, thanks.

build-push/bin/main.sh
cevich: Maybe, I can move it.  And it seems disabling SC2154 here isn't needed anyway, so I'll nuke it.

cevich:
> And... sorry again, but I can't really review this. It's huge, and there's a lot of outside context I'm not privy to.

Yep I understand.  Most of the outside context you've reviewed in the past, but it is complex overall.  However, my goal here is to re-implement [the (3x duplicated) github-action workflow](https://github.com/containers/podman/blob/a08e4e5b4657469a5d686ec5621f681cd240f6e7/.github/workflows/multi-arch-build.yaml) in a way that's unit & integration tested, can be re-used across all three repos.,  and is mostly human readable/maintainable (given some amount of "study-time" of course).

Though I totally understand reviewing this likely also requires at least some understanding of the github-action workflow...which is a built-up mess, tweaked and extended based on lots of group discussions :confounded:  Perhaps it would be helpful to future maintainers, if instead of one big `main.sh`, I broke this up further into three scripts:  One for handling the 'stable' image, and two more for dealing with 'testing' and 'upstream' images?

Some Context: These images are ultimately downloaded thousands of times per month, so if some change is needed in the build process, I want this overall build setup to be maintainable by people other than me.  Maybe that's asking too much though, I can't tell.  If you have readability suggestions, I'm definitely all ears :ear:

cevich:
Converting this back to draft-status so the `DO NOT MERGE` commit doesn't accidentally get merged in.

README.md
```diff
@@ -99,6 +99,12 @@ see step 4 below.
 * The bulk of the packaging work occurs next, from the `cache_images/*_packaging.sh`
   scripts.  **This is most likely what you want to modify.**
 
+* Unlike the Fedora and Ubuntu scripts, the `build-push` scripts (and VM image)
+  is not for general-purpose use.  It's tightly-coupled with `build-push` scripts
+  in the
+  [containers/automation repository](https://github.com/containers/automation).
+  Chances are good that's what you want to change, not the VM image or scripts.
+
```
cevich, line 107: This is confusing / sucks.  I'll fix it.

cevich:
Force-push: removed temporary `DO NOT MERGE` commit.

edsantiago:
LGTM but the github diff page is showing a bunch of ugly dark yellow warning messages

cevich:
> a bunch of ugly dark yellow warning messages

Ya, ignore those, they're false-positive YAML problems.  Anyway, thanks for looking, I'll fix up the docs and get this in.

PR #104, opened 2021-12-08
[CI:TOOLING] Build tooling containers in parallel

cevich:
This build takes 17-20min when run sequentiually.  Make use of `make`
jobs option to do the builds in parallel.  Then rely on the
`ci/make_container_images.sh` script just for it's push logic.


cevich:
Closing, shared DNF cache is causing conflicts with parallel builds accessing it.

PR #111, opened 2022-01-26
Multiple pythons needed for podman-py testing

cevich:


jwhonce:
@cevich I don't know, the cirrus.yaml has a container field. Does that mean the tests are only run in a container?

cevich:
Sorry, I thought maybe you'd know off-hand.  Taking a look in podman-py, I see the `gating` task uses the container.  Both `test` and `latest` use VMs, so we do have a mix :cry:  Assuming `make` and `make lint` need to use those other python versions, my idea of asymmetrically installing won't work.  Damn.

cevich:
Let's try it this way: I moved most of the python bits into the VM-only list of installs.  We'll need to confirm the images work with podman CI, but this will all but negate any container-image size concerns (unless it doesn't work)...

cevich:
K, lessee if it works in podman.  You're welcome to test `c6730261319647232` in podman-py CI as well.

jwhonce:
@cevich See https://github.com/containers/podman-py/pull/160/checks?check_run_id=4973220410 for errors

cevich:
```
make
rm dist/* || :
rm: cannot remove 'dist/*': No such file or directory
/usr/bin/python3 -m pip install --user -r requirements.txt
/usr/bin/python3: No module named pip
make: *** [Makefile:16: podman] Error 1

Exit status: 2
```

Right, okay so we need some python stuffs in the containers then.  Lemmie go into one manually where I can `dnf install` bit by bit until the tests pass...

cevich:
...oof, okay, no won't work.  `make lint` wants all the bunches of python versions too :disappointed:  Probably the quickest way around this is to alter the `gating` CI job to use VMs and not containers.  That will slow the job down a little, since the VM has to boot, but will not result in bloat or extra things to maintain on the image-side.

That said (and I remember we talked about this before), it seems like we need a custom image for the `podman-py` repo. CI.  The python needs are unique enough that the one-size-fits-all approach shows weakness.  This is more/less why I did custom images for `netavark` and `aardvark`.

This does take quite a bit of effort to pull off, and we're in the middle of the. podman 4.0 crunch.  Is a working  https://github.com/containers/podman-py/pull/160 necessary/critical for the 4.0 efforts?

I can spend a little time on it today, since other folks are OOO - and I'm blocked on their help to move some items forward....

cevich:
The `fedora-podman-py Cache Image` is the CI task to watch.  Assuming it builds successfully, the hard-work can begin: getting it to work in the `podman-py` repo...

cevich:
Repos timed out, retrying...

cevich:
`Timed out!` ugh...it must be Friday.  I think the repos. (or network) are just being slow today for whatever reason.  I'll try again Monday.

cevich:
Still timing out.  I don't get it.  The other VM images build in < 10 minutes.  WTH going on with the new one it taking so long.

cevich:
Found the problem...there was a `dnf install ...` command waiting for somebody to press `y` :sob:

cevich:
AAAAAND if it ain't one thing...it's something else...and something else...and...

cevich:
Looks like I pulled too many python packages out from the main image ([log](https://cirrus-ci.com/task/4695080597454848?logs=setup#L206-L207))...

cevich:
***DO NOT USE***: `c5533341981081600` - has netavark/aardvark packages installed for F35 by default.

PR #114, opened 2022-02-16
Trigger build of fresh VM  images for podman 4.0

cevich:
This is to ensure VM images for CI, which contain the
intended dependency versions to support the podman
4.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ see step 4 below.
   "extra" code, remember these VM images are shared by automation in multiple
   repositories.
 
+
 ### Process: ###
 
 1. After you make your script changes, push to a PR.  They will be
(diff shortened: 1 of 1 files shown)


cevich:
Ref: https://docs.google.com/document/d/1jP-BY37-lGX3IMx4HHEsNiS_4Upw59tD2fevmpYrDrQ/edit

cevich:
Built container images (nearly identical to VMs):
* `quay.io/libpod/fedora_podman:c5814666029957120`
* `quay.io/libpod/prior-fedora_podman:c5814666029957120`
* `quay.io/libpod/ubuntu_podman:c5814666029957120`

cevich:
@rhatdan @lsm5 looks like we need F35 version bumps (if possible) or updates to the google doc (linked above) for:
* container-selinux
* fuse-overlayfs
* slirp4netns

Note: The point here is that versions in CI matches (or exceed) what we plan to ship in RHEL, to help catch any "surprises" before they make it to gating tests and/or users

PR #116, opened 2022-03-03
Netavark: Install rust from upstream

cevich:
The Fedora rust toolchain is fairly old, especially for a young
language.  It also does not include the rustup utility which is
hampering netavark/aardvark-dns development.  Rather than wait for
Fedora to catch up, install the upstream toolchain and utilities.

### DIFF
--- a/cache_images/fedora-netavark_packaging.sh
+++ b/cache_images/fedora-netavark_packaging.sh
@@ -23,8 +23,6 @@ INSTALL_PACKAGES=(\
     bind-utils
     bridge-utils
     bzip2
-    cargo
-    clippy
     curl
     dbus-daemon
     findutils
@@ -45,11 +43,7 @@ INSTALL_PACKAGES=(\
     podman
     policycoreutils
     podman
-    redhat-rpm-config
-    rpm-build
     rsync
-    rust
-    rustfmt
     sed
     slirp4netns
     tar
@@ -59,14 +53,16 @@ INSTALL_PACKAGES=(\
 )
 
 # TODO: Remove this when all CI should test with Netavark/Aardvark by default
-EXARG="--exclude=netavark --exclude=aardvark-dns"
+EXARG="--exclude=netavark --exclude=aardvark-dns --exclude=cargo --exclude=rust"
 
 msg "Installing general build/test dependencies"
 bigto $SUDO dnf install -y $EXARG "${INSTALL_PACKAGES[@]}"
 
-msg "Installing netavark-specific toolchain dependencies"
-export CARGO_HOME="/var/cache/cargo"  # must match .cirrus.yml in netavark repo
-$SUDO env CARGO_HOME=$CARGO_HOME cargo install mandown cross
-
 # It was observed in F33, dnf install doesn't always get you the latest/greatest
 lilto $SUDO dnf update -y
+
+msg "Initializing upstream rust environment."
+export CARGO_HOME="/var/cache/cargo"  # must match .cirrus.yml in netavark repo
+$SUDO mkdir -p $CARGO_HOME
+# CI Runtime takes care of recovering $CARGO_HOME/env
+curl https://sh.rustup.rs -sSf | $SUDO env CARGO_HOME=$CARGO_HOME sh -s -- -y


cevich:
FYI: @rhatdan this PR will also get you new images for podman w/ updated crun, runc, etc.

flouthoc:
LGTM. I am ok with VM using the upstream toolchain for rust, with this in place we can remove lot of custom pulling from netavark's CI code.

@cevich but does this also puts us at a risk that we are not testing with what gets shipped with `dnf` ? However concerned about this a lot cause if it builds with one toolchain it is highly likely that it will also build with the older one.

cevich:
You're concerns are valid, and it's not my first-choice.  In my experience it does cause problems, especially in packaging.  Hopefully they can/will be caught by downstream gating tests (assuming we even have them).  In any case, you're right, and it is somewhat "kicking the can down the road" for somebody else to deal with.  At this point though, we really don't know what we don't know, so I say we give it a try.

flouthoc:
@cevich Sorry I did a type in my above comment. I am also not concerned a lot about this since its unlikely that builds will fail with what's on `fedra` dnf.

So yes i agree lets give it a try.

cevich:
```
fedora-netavark: sudo: showrun: command not found
```

Oops.  Trying again...

cevich:
@rhatdan & @flouthoc give `c5309576894480384` a try.

flouthoc (commented):
Testing new image here: https://github.com/containers/netavark/pull/252

flouthoc:
@cevich current `rustup-init` does not configures `rustc` profile correctly something like `curl https://sh.rustup.rs -sSf | sh -s -- -y` does that correctly but its not mandatory I can do that in `build_cross` task specifically.

Other than above image is working correctly and also contains `podman` as well, I don't think any change is mandatory I could do that in `setup.sh` so its not a problem

cevich:
Force-push: Rebuilding to include configuring `rustc`, def. something we want to do here and not at test runtime.

flouthoc:
Pushed a change with new image.

cevich:
>Pushed a change with new image.

So nothing meaningful actually happens when I merge this, the images are already built.  So I'll hold off until https://github.com/containers/netavark/pull/252 merges, just in case any further related changes are needed.

cevich:
[There's an error that's being ignored](https://cirrus-ci.com/task/6676947219513344?logs=main#L667-L675) but looks bad.  And it seems like I left the `cargo` package installed on accident...

cevich:
...and the curl command was messed up.  All my fault.  Building again...

flouthoc:
@cevich Did you forgot git add, I still can't see the part where `fedora`'s rust is being removed.

cevich:
> I still can't see the part where fedora's rust is being removed.

Oh, I just removed `rust` and `cargo` from the `$INSTALL_PACKAGES` list.  Ahh, I see another problem though, the `$CARGO_HOME` env. var. is being blocked by `sudo`...

cevich:
force-push: Get `$CARGO_HOME` through `$SUDO` (script runs as a temporary/regular user, not root).

cache_images/fedora-netavark_packaging.sh
```diff
@@ -64,9 +61,17 @@ EXARG="--exclude=netavark --exclude=aardvark-dns"
...
+export CARGO_HOME="/var/cache/cargo"  # must match .cirrus.yml in netavark repo
+$SUDO mkdir -p $CARGO_HOME/bin/
+cd $CARGO_HOME/bin/
+$SUDO curl --fail --silent --location -O --url \
+    "https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init"
+$SUDO chmod +x rustup-init
+# CI Runtime takes care of recovering $CARGO_HOME/env
+$SUDO env CARGO_HOME=$CARGO_HOME ./rustup-init -y --no-modify-path
+# Configure the sysroot for rustc here, so it remains consistent at test runtime
+curl https://sh.rustup.rs -sSf | $SUDO env CARGO_HOME=$CARGO_HOME sh -s -- -y
```
cevich, line 77: Hmmm, looking at `https://sh.rustup.rs` it appears to duplicate running rustup-init.  There's no reason to do this twice.

cevich:
force-push: Only run `rustup-init.sh` once :grin:

cevich:
Ugh...somehow:

```
    ...cut...
    fedora-netavark:   Installing       : rust-std-static-1.58.1-1.fc35.x86_64                79/110
    fedora-netavark:   Installing       : rust-1.58.1-1.fc35.x86_64                           80/110
    fedora-netavark:   Installing       : cargo-1.58.1-1.fc35.x86_64                          81/110
    ...cut...
```

Ahh, it's rpm build that's pulling it in.  Removing that, and also adding rust/cargo to the exclusion list to make sure.

cevich:
Force-push: Removed `clippy` due to

```
�[0;32m    fedora-netavark: Error:�[0m
�[0;32m    fedora-netavark:  Problem: conflicting requests�[0m
�[0;32m    fedora-netavark:   - package clippy-1.55.0-1.fc35.x86_64 requires rust(x86-64) = 1.55.0-1.fc35, but none of the providers can be installed�[0m
```

cevich:
There we go, [this is what we want to see](https://cirrus-ci.com/task/4627410124537856?logs=main#L471).

flouthoc:
@cevich New image is working fine. Just missing some `build tools` and `gcc`. Could you add this `dnf -y install make automake gcc gcc-c++ kernel-devel` and this is good to go.

cevich:
> New image is working fine. Just missing some build tools and gcc. Could you add this dnf -y install make automake gcc gcc-c++ kernel-devel and this is good to go.

Yep, I'll do that in a separate PR.  No need to rebuild images for that small change.

PR #121, opened 2022-03-30
[CI:DOCS] Add get_ci_vm warning about ssh key passphrases

cevich:
When initializing, gcloud will check for a special ssh key under
~/.config/gcloud/ssh.  When absent, it will generate a new one and
prompt the user for a passphrase.  If a passphrase is set, it will be
requested again for every single gcloud operation, with no possibility
to add it to a desktop keychain.  To help prevent this aggravation,
include it in a warning message at initialization time.


baude:
lgtm

PR #125, opened 2022-04-07
Build-push: Also tag vX.Y and vX stable images

cevich:
Since these images build daily, and the full (X.Y.Z) version could
change at any time, it may be difficult for downstream users to track a
uniform FQIN while also receiving security updates.  Since all tags of
produced manifest-lists are already pushed, simply update the
tag_version.sh mod-script to produce the additional tagged
manifest-lists.  Verify this has happened by enhancing the CI test to
check for the new tags.

### DIFF
--- a/build-push/bin/tag_version.sh
+++ b/build-push/bin/tag_version.sh
@@ -33,11 +33,26 @@ if [[ "$CTX_NAME" == "stable" ]]; then
     # retrieving it from the image content is beyond the scope
     # of this script.
     req_env_vars img_cmd_version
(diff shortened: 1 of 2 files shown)


cevich:
@TomSweeneyRedHat and @edsantiago this will only affect the `stable` images of Podman.  I want to make sure this replacement for the old github-action is working well before replacing it in Skopeo and Buildah.

TomSweeneyRedHat:
NIce!
TYVM @cevich

PR #129, opened 2022-04-22
Support building either mainline or beta GCE images based on PR label

cevich:
Supporting a new base-image in downstream environments is often
a time-consuming and difficult.  For Fedora, normally this work begins
when a release goes into beta.  However, beta versions of Fedora
releases are not provided in google, they must be specially imported.
Otherwise, after release the `fedora-cloud` project provides read-made
images that can be used in a vastly simpler manner.

Alter the base-image build workflow to use the pre-built images by
default.  However if a PR has the 'Beta' label set, cause the workflow
to import the development fedora images instead.


cevich:
Ref: [Successful build with 'Beta' tag.](https://cirrus-ci.com/build/5833015967350784)

cevich:
@edsantiago and @lsm5 would you mind taking a glance at this.  Especially the `README.md` updates.  I don't expect y'all to grok the YAML or script changes, and [they both work as intended.](https://github.com/containers/automation_images/pull/129#issuecomment-1116260129).  This isn't an emergency or critical fix, just intended to simplify the workflow for the general (i.e. non-beta) image-update use-case.

lib.sh
```diff
@@ -98,6 +99,55 @@ set_gac_filepath(){
...
+    result=$(curl --location \
+             -H "$h_auth" -H "$h_accept" -H "$h_content" \
+             --request POST --data @- --url "$api" <<<"$query") \
+             || \
+             die "Error communicating with GraphQL API $api: $result"
+
+    # GraphQL sometimes returns errors inline, assume no result
+    # format or data should ever contain the string 'error'
+    if grep -q "error" <<<"$result"; then
+        die "Found 'error' in result: $result"
```
edsantiago, line 140: Cringe: if anyone ever adds a label named 'ignore-errors' or 'terrorific', this will break in hard-to-diagnose ways. Could you check instead for an empty list in the `jq` below?

base_images/fedora_base-setup.sh
```diff
@@ -57,8 +57,9 @@ if ! ((CONTAINER)); then
         # vs google-network-daemon.service.  Fix this with a custom
         # cloud-init service file.
         CLOUD_SERVICE_PATH="systemd/system/cloud-init.service"
-        echo "$sourcemsg" > /etc/$CLOUD_SERVICE_PATH
-        cat $SCRIPT_DIRPATH/fedora-cloud-init.service >> /etc/$CLOUD_SERVICE_PATH
+        echo "$sourcemsg" | $SUDO tee -a /etc/$CLOUD_SERVICE_PATH
```
edsantiago, line 60: This is not equivalent: the old code clobbers `$CLOUD_SERVICE_PATH`, the new code appends. If I were a future maintenance programmer, I would see the `-a` and immediately start looking for when/where/who creates and initializes that file. I would encourage you to drop the `-a` here and in the other instance below.

And, taking a step back, I don't understand the reason for the `tee`. Is it so the contents are visible in a log somewhere? If so, it doesn't seem very helpful, because it's just undescribed content. Maybe an 'echo this is what $CLOUD_SERVICE_PATH will look like'?

edsantiago (changes requested):
Concerned about the 'error' grep and about `tee -a`. 

The YAML is taking me a long time to review; looks fine on first (diff) and second (once-over) pass, and I don't think I have the ability to give it another look today.

base_images/fedora_base-setup.sh
cevich: > This is not equivalent

Oh woops, good catch.

> I don't understand the reason for the tee

The previous build process had root access to run the script, the new build process *sometimes* has root access.  The `$SUDO tee` guarantees the file can always be created.

cevich:
> The YAML is taking me a long time to review; looks fine on first (diff) and second (once-over) pass, and I don't think I have the ability to give it another look today.

Please don't fret too much over the YAML, it's special and specific to Packer, so I really don't expect it to be widely understandable.  That said, if you're willing to put in the effort and learn more about that tool, I'm not going to get in the way.  I will note that the docs for 1.7 have some bad links use 1.8, esp. for [the plugins](https://www.packer.io/plugins).

cevich:
I'm going to fork off the `lib.sh` addition that handles reading PR labels.  I can use that feature in other PRs, whereas the support for consuming upstream GCP F36 images has hit a speed-bump: https://pagure.io/cloud-sig/issue/384

cevich:
[There's been no activity on my upstream issue for over a week](https://pagure.io/cloud-sig/issue/384).  I'm taking that as a sign the Fedora project doesn't take maintenance of these GCE images seriously.  And even if so, they're not updated quickly enough for our CI purposes.

PR #130, opened 2022-04-26
[CI:DOCS] Minor: Update NM workaround for netavark

cevich:
While the `veth*` is probably the effective statement for preventing
CI test flakes, also ignore any bridges/interfaces with `podman` in
the name.


cevich:
Note to me: Waiting for F36 PR to merge.

PR #133, opened 2022-05-24
Download/cache 'podman-plugins' package in Fedora

cevich:
Specifically, the 'dnsname' CNI plugin is needed by CI testing in
containers/common.  Download it into the VM for installation at runtime.


cevich:
Oof, we've reached Ubuntu 21.10 EOL, so I can't build new images here.  We'll need to wait for support to be added.

PR #138, opened 2022-06-14
[CI:TOOLING] Read pr labels

cevich:
On occasion it's necessary to alter automation behavior based on some
criteria.  For high-level conditions, using magic-words in the PR title
is quick/dirty but somewhat cumbersome.  Using the `required_pr_labels`
mechanism in `.cirrus.yml` is handy too, unless the 'absence' of a label
needs to be tested.  Add a generic library function that allows retrieving
the labels on a PR.  Future commits can make use of this in scripts to
take various actions based on presence or absence of the labels.


cevich:
@edsantiago PTAL (you've seen this code before).  I separated it from #129 since the status of GCP images is questionable, but I need to read PR labels for another/unrelated PR.  Thanks in advance.

ci/Containerfile
```diff
@@ -40,6 +41,13 @@ RUN dnf update -y && \
     dnf autoremove -y && \
     dnf clean all
 
+# There's no simply way to install a automation-library version that's
+# consistent with other environments via use of the ../lib.sh specification.
+# Just use the "latest" version and assume it will work for this containers context.
+RUN url='https://github.com/containers/automation/releases/latest/download/install_automation.sh' && \
```
edsantiago, line 47: Why is this different from the URL in `lib.sh`? https://github.com//containers/automation_images/blob/841849cd3e9e301f202ab886f5cc658d9a21dc98/lib.sh#L57

And, did you test it? (Is it possible to test?) I don't understand Containerfile syntax enough to know if this `var=xxx && so on` syntax will work.

And, it really disturbs me to have this sort of moving-target usage... but since it's already being done in lib.sh, I guess we live with it.

ci/Containerfile
```diff
@@ -40,6 +41,13 @@ RUN dnf update -y && \
     dnf autoremove -y && \
     dnf clean all
 
+# There's no simply way to install a automation-library version that's
```
edsantiago, line 44: minor nit: should be `simple`

image_builder/install_packages.sh
```diff
@@ -34,3 +34,5 @@ if (("${CONTAINER:-0}")); then
     dnf mark install dnf yum $(<"$INST_PKGS_FP")
     dnf autoremove -y
 fi
+
+install_automation_tooling
```
edsantiago, line 38: I don't see anything in the commit message about why this was added

cevich: Good catch...now I'm wondering that myself.  It may have leaked in from the other PR.  I'll check and fix the commit message or remove it.

cevich:
Force-push: Updates from review suggestions.  [The validate tasks "pre-build container image" dependency will confirm image changes are functional.](https://cirrus-ci.com/task/6370438765871104)

ci/Containerfile
edsantiago: Thanks for fixing it. Final (I think) question: since one of these is in a Containerfile, and the other in a .sh file, is there any way that it could be refactored so the duplication is eliminated? (I couldn't figure out where `ci/Containerfile` gets run, but maybe there's a way for the caller to set the URL via envariable?)

cevich: > maybe there's a way for the caller to set the URL via envariable?

Actually that gives me an idea, I think this might be possible...

cevich: ...force-pushed something that might work.

cevich:
Changed to draft so I remember to squash the `fixup!` commit.

ci/Containerfile
cevich: [Yep it works!](https://cirrus-ci.com/task/5036839777075200?logs=build#L1123)

cevich:
force-push: squashed

edsantiago:
Cleverly solved, thank you!

/lgtm

cevich:
Thanks for the help reviewing!

Issue #139, opened 2022-06-20
install bats directly into the VM images


Luap99:
```
VERSION=v1.1.0 ./hack/install_bats.sh
Cloning into '/tmp/tmp.WsaMSX61iB'...
fatal: unable to access 'https://github.com/bats-core/bats-core/': Failed to connect to github.com port 443: Connection timed out
make: *** [Makefile:842: .install.bats] Error 128
```
From one of my PRs: https://cirrus-ci.com/task/4733813895987200

To prevent flakes bats should be in the VM images and not installed at runtime.

bats should available in both fedora and debian repos,
on fedora v1.7.0 is available and on ubuntu (21.10/22.04) v1.2.1 https://packages.ubuntu.com/search?keywords=bats

Given that the CI script installs v1.1.0 for our testing using the packaged version should just work?

@edsantiago @cevich PTAL

edsantiago:
We can't rely on system-provided bats in Ubuntu: https://bugs.launchpad.net/ubuntu/+source/bats/+bug/1882542

However, yes, obviously we should install it at VM-create time, never runtime. Network access at runtime must always be the absolute minimum.

cevich:
Agreed, and [actually we already do this.](https://github.com/containers/automation_images/blob/main/cache_images/podman_tooling.sh#L31).  Though it's coming in from podman's `Makefile`, which...[ahh here's the bug.](https://github.com/containers/podman/blob/main/hack/install_bats.sh#L7)  I'm pretty sure that should be `if [[ "$(type -P bats)" != "" ]]; then`.  Also, updating the make target to specify the hard-coded `/usr/local/bin/bats` as a dependency will help avoid unnecessarily calling the script.  I can toss up a PR real quick...

Luap99:
What version is installed in the VM? I think it would be great to upgrade to something more recent, if I remember correctly last time I used hack/get_ci_vm the bats version didn't have the `--filter` feature which is very hand when you only need to run a single test.

cevich:
The bats package isn't installed in [fedora](https://github.com/containers/automation_images/blob/main/cache_images/fedora_packaging.sh) or [ubuntu](https://github.com/containers/automation_images/blob/main/cache_images/ubuntu_packaging.sh).  Instead, both clone the podman repo and call `make install.tools`.

Given @edsantiago (reasonable) objection to https://github.com/containers/podman/pull/14686 I'm thinking:
[...]

PR #140, opened 2022-06-22
Use packaged bats in Fedora, side-load for Ubuntu

cevich:
Fixes https://github.com/containers/automation_images/issues/139

N/B: Using these images requires an update to the podman repo. to remove
the bats installer script and `Makefile` target.

### DIFF
--- a/cache_images/podman_tooling.sh
+++ b/cache_images/podman_tooling.sh
@@ -21,7 +21,7 @@ if [[ "$OS_RELEASE_ID" == "ubuntu" ]]; then
     bats_version="1.7.0"
     dl_url="https://github.com/bats-core/bats-core/archive/v${bats_version}.tar.gz"
     echo "Installing bats $bats_version"
-    curl --fail --location "$dl_url" | tar xzv -C /tmp
+    curl --fail --location "$dl_url" | tar xz -C /tmp
     pushd /tmp/bats-core-$bats_version
     $SUDO ./install.sh /usr/local  # prints install location
     popd
--- a/cache_images/podman_tooling.sh
+++ b/cache_images/podman_tooling.sh
@@ -6,13 +6,27 @@
 # as root or through sudo.  The script should not be used for any other
 # purpose or from any other context.
 
-set -e
+set -eo pipefail
 
 SCRIPT_FILEPATH=$(realpath "${BASH_SOURCE[0]}")
 SCRIPT_DIRPATH=$(dirname "$SCRIPT_FILEPATH")
 REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
 # shellcheck source=./lib.sh
 source "$REPO_DIRPATH/lib.sh"
+
+if [[ "$OS_RELEASE_ID" == "ubuntu" ]]; then
+    if [[ -n "$(type -P bats)" ]]; then
+        die "Bats _MUST_ not be installed on ubuntu until fixed: https://bugs.launchpad.net/ubuntu/+source/bats/+bug/1882542"
+    fi
+    bats_version="1.7.0"
+    dl_url="https://github.com/bats-core/bats-core/archive/v${bats_version}.tar.gz"
+    echo "Installing bats $bats_version"
+    curl --fail --location "$dl_url" | tar xzv -C /tmp
+    pushd /tmp/bats-core-$bats_version
+    $SUDO ./install.sh /usr/local  # prints install location
+    popd
+    rm -rf /tmp/bats-core-$bats_version
+fi
 
 echo "Configuring Go environment"
 export GOPATH="${GOPATH:-/var/tmp/go}"
--- a/cache_images/fedora_packaging.sh
+++ b/cache_images/fedora_packaging.sh
@@ -41,6 +41,7 @@ INSTALL_PACKAGES=(\
     autoconf
     automake
     bash-completion
+    bats
     bridge-utils
     btrfs-progs-devel
     buildah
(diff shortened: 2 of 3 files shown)


cevich:
Crossing my fingers and hoping this works (it may be slightly over-complicated) :stuck_out_tongue_winking_eye:

cache_images/podman_tooling.sh
```diff
@@ -14,6 +14,23 @@ REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
...
+if [[ "$OS_RELEASE_ID" == "ubuntu" ]]; then
+    if [[ -n "$(type -P bats)" ]]; then
+        die "Bats _MUST_ not be installed on ubuntu until fixed: https://bugs.launchpad.net/ubuntu/+source/bats/+bug/1882542"
+    fi
+    req_env_vars UBUNTU_BATS_VERSION
+    # shellcheck disable=SC2154
+    echo "Installing bats $UBUNTU_BATS_VERSION to /usr/local"
+    buildDir=$(mktemp -d)
+    lilto git clone -b "$UBUNTU_BATS_VERSION" \
+        https://github.com/bats-core/bats-core $buildDir
```
edsantiago, line 26: I think I'd be a lot happier with:
```bash
  wget https://github.com/bats-core/bats-core/archive/v${UBUNTU_BATS_VERSION}.tar.gz
  tar -C ... xvf    <--- goes into bats-core-x.y.z
```
Do you have a strong reason for doing it via git?

podman/Containerfile
```diff
@@ -22,7 +24,9 @@ ADD /base_images/* "$AI_PATH/base_images/"
 ADD /cache_images/* "$AI_PATH/cache_images/"
 ADD /systemd_banish.sh "$AI_PATH/"
 WORKDIR "$AI_PATH"
-RUN bash ./podman/setup.sh
+RUN cd $AI_PATH && \
+    chmod +x ./podman/setup.sh && \
+    $AI_PATH/podman/setup.sh
```
edsantiago, line 29: Is the mix-and-match of absolute and relative paths necessary? `s/$AI_PATH/./` in line 29 might make it more readable.

edsantiago (commented):
Well, you're right, that was a lot more than I'd expected :)

Two concerns (not blockers, but more than nits). And a caveat that much of this (the gce.yml) is way above my head. But, yeah, LGTM. Thanks for the quick work.

podman/Containerfile
cevich: Oh woops, I intended to fix that.  Thanks for pointing it out.

cevich:
> And a caveat that much of this (the gce.yml) is way above my head.

I've stripped that out and vastly simplified things.  Then I went and screwed up again by including a commit I should just make a separate PR for.  Sigh...there's always time for another try...

cevich:
...alright, fingers crossed.

cevich:
[The 'ubuntu Cache Image' build failure](https://cirrus-ci.com/task/5153800628469760?logs=main#L134) seems to be an upstream problem:

```
    ubuntu:  libsystemd-dev : Depends: libsystemd0 (= 249.11-0ubuntu3) but 249.11-0ubuntu3.3 is to be installed
    ubuntu:  libudev-dev : Depends: libudev1 (= 249.11-0ubuntu3) but 249.11-0ubuntu3.3 is to be installed
```

[At least for libsystemd-dev, it's reported here.](https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1979579)  I'm assuming once this is fixed, the libudev1 package will get fixed as well.

edsantiago:
Passed on re-run. I still haven't done a final review, but in my first passes (on earlier revisions) all seemed fine.

cevich:
> Passed on re-run.

Oh, thanks a bunch for re-running that!  So it was def. a hangup in the upstream repos.  Anyway, no rush to review, I can get a PR going in podman to test the new image (and remove the `install_bats.sh`) with this one still open.  But I fully expect that PR to fail, since I think Lokesh's previous image-update PR is still running into problems.  But we can at least make sure bats is found and starts up...

cache_images/podman_tooling.sh
```diff
@@ -6,14 +6,28 @@
...
 source "$REPO_DIRPATH/lib.sh"
 
+if [[ "$OS_RELEASE_ID" == "ubuntu" ]]; then
+    if [[ -n "$(type -P bats)" ]]; then
+        die "Bats _MUST_ not be installed on ubuntu until fixed: https://bugs.launchpad.net/ubuntu/+source/bats/+bug/1882542"
+    fi
+    bats_version="1.7.0"
+    dl_url="https://github.com/bats-core/bats-core/archive/v${bats_version}.tar.gz"
+    echo "Installing bats $bats_version"
+    curl --fail --location "$dl_url" | tar xzv -C /tmp
```
edsantiago, line 24: After reading the [ubuntu log](https://cirrus-ci.com/task/6658219291443200?logs=main#L1711) I belatedly realized that the `-v` is a bit excessive... but not worth a re-push over. Will lgtm-hold so you can test the images before merging. Thanks @cevich 

/lgtm
/hold

edsantiago: Oh, this repo doesn't work that way :-). LGTM anyway.

cevich: Sounds good, if this PR needs re-pushing for new images, I can fix that up quick.

cevich:
Buildah PR: https://github.com/containers/buildah/pull/4074

cevich:
Rebuilding images due to [problems with podman package in Ubuntu](https://github.com/containers/podman/pull/14397#discussion_r908852677).

cache_images/podman_tooling.sh
cevich: fixed.

PR #141, opened 2022-06-22
[CI:TOOLING] Fix failure to upload package cache

cevich:
Fix directory name and required upload argument problem in
`.cirrus.yml`.  Also fix container images such that they actually retain
their `dnf` or `apt` cache (volume-mounted from host).  This should
provide a minor build speedup over time.


cevich:
note to me: I don't think Cirrus is caching all the subdirectories, maybe try adding a wildcard to the path.

cevich:
Note to me: This isn't working, [multiple runs of tooling image build didn't get any faster](https://cirrus-ci.com/task/5445218467053568).

cevich:
Ugh, this is too hard/complex.  Giving up for now (maybe forever).

PR #143, opened 2022-06-30
Skip specified image builds based on PR labels

cevich:
With an ever growing number of images to build, it's inevitable that
a flake will occur on an image the author does not care about.  Should
this occur, provide a runtime-method by which a 'no_<name>' label can be
added to the PR before re-running the failure.  The specified `<name>`
build will then be bypassed.

Note: This will not bypass container tooling images (they're always
required). Nor will it properly handle any build dependencies.
For example a `no_fedora` label will cause a build failure of the
`build-push` image, since it depends on the `fedora` base image.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -247,6 +247,8 @@ imgts_task:
         cpu: 2
         memory: '2G'
     env:
+        # Not all $IMGNAMES may have been built, don't fail if some are missing
+        REQUIRE_ALL: 0  # This should ONLY be set `0` on this repository!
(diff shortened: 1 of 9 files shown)


cevich:
Fudge...need workaround for `Apply new image metadata` :disappointed:

cevich:
Okay, this is working as intended and properly issuing a [warning on timestamp update failures.](https://cirrus-ci.com/task/4832911085010944?logs=main#L22) instead of failing outright.

PR #147, opened 2022-07-12
[WIP]Add containers org link

baude:
Add link to containers repo.  But really, need to build new images with
fresh packages.

Signed-off-by: Brent Baude <bbaude@redhat.com>


edsantiago:
LGTM but reminder that these generated images *CAN NEVER BE USED IN PODMAN CI*.

edsantiago:
Wait - why close? The change is a good one, I think it would be good to merge, I just meant that we need to make sure we never open a podman PR that uses these images.

PR #149, opened 2022-07-14
Fix dnf update installing netavark/aardvark-dns

cevich:
These packages should _never_ be pre-installed in the CI images for
testing in the netavark & aardvark-dns repositories.  Ensure they aren't
brought in by a package update in addition to the install.

### DIFF
--- a/cache_images/fedora-netavark_packaging.sh
+++ b/cache_images/fedora-netavark_packaging.sh
@@ -61,16 +61,26 @@ INSTALL_PACKAGES=(\
     zip
 )
 
-EXARG="--exclude=netavark --exclude=aardvark-dns --exclude=cargo --exclude=rust"
+EXARG="--exclude=cargo --exclude=rust"
 
 msg "Installing general build/test dependencies"
 bigto $SUDO dnf install -y $EXARG "${INSTALL_PACKAGES[@]}"
 
-# It was observed in F33, dnf install doesn't always get you the latest/greatest
-lilto $SUDO dnf update -y
+# It was observed in F33, dnf install doesn't always get you the latest/greatest.
+lilto $SUDO dnf update -y $EXARG
 
 msg "Initializing upstream rust environment."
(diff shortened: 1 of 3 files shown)


cevich:
Ref: https://github.com/containers/netavark/pull/308 & https://github.com/containers/netavark/pull/344

cevich:
force-push: Fixed permission-denied problem.

flouthoc (approved):
LGTM

flouthoc:
I see some failures in `build_cross` with this image: https://cirrus-ci.com/task/6066640541450240

lsm5:
shouldn't we pull packages from podman-next copr instead?

lsm5:
I'm thinking this:
- Install netavark, aardvark, podman etc. from the podman-next COPR. Do not remove any files here.
- In the aardvark CI env, `rm /usr/libexec/podman/aardvark-dns`
- In the netavark CI env, `rm /usr/libexec/podman/netavark`

That way we can perhaps get rid of build_cross targets altogether and also not worry about curling zips or tarballs.

lsm5:
hmm, installing packages here would mean the image would need to get built continuously to test with the latest commits merged upstream? Maybe any package installation from the copr should actually be done in the individual repos and not here?

cevich:
> Maybe any package installation from the copr should actually be done in the individual repos and not here?

Yes, and good on you for observing this.  Yes, we want the stuff being tested to come from builds in the PR and/or be very recent builds from podman-next (for dependencies).  Otherwise, for the basic environment stuff that doesn't need to be updated constantly, it should be pre-installed in the images.

cevich:
Ugh, I need to go manually trudging through the tasks and pull out image Id's, need to update that bot.  So looks like:

* GCE image ID is `c5353795591864320`
* fedora-aws is `ami-09e5978f141900855`
* fedora-netavark-aws-arm64 is `ami-00d7ccde5eaacc50e`
* fedora-podman-aws-arm64 is `ami-02da46d8fa51e5316`

PR #154, opened 2022-07-18
Makefile: build fedora_podman for both amd64 and arm64

lsm5:
Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

@cevich PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,6 +238,7 @@ $(_TEMPDIR)/%_podman.tar: podman/Containerfile podman/setup.sh $(wildcard base_i
 		--build-arg=BASE_NAME=$(subst prior-,,$*) \
 		--build-arg=BASE_TAG=$(call err_if_empty,BASE_TAG) \
 		--build-arg=PACKER_BUILD_NAME=$(subst _podman,,$*) \
+		--platform linux/amd64,linux/arm64 \
 		-v $(_TEMPDIR)/.cache/$*:/var/cache/dnf:Z \
 		-v $(_TEMPDIR)/.cache/$*:/var/cache/apt:Z \
 		-f podman/Containerfile .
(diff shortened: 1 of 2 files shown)


lsm5:
i'm hoping the failed jobs were just a timing mismatch between submitting PR and adding labels.

lsm5:
@cevich looks like `qemu-user-static` is the missing piece that's causing the arm64 failure. I'm still looking for the right repo on c8s that'll get me that package

lsm5:
@cevich and looks like `qemu-user-static` is not shipped on c8s. So, I guess it's either c9s or fedora at this point, wdyt?

nalind:
You're probably going to want to replace `-t $*_podman:$(call err_if_empty,IMG_SFX)` with `--manifest $*_podman:$(call err_if_empty,IMG_SFX)` to put the built images into a manifest list instead of trying to tag both of them with the same name, which isn't possible.  I don't know if `podman save` will handle that, though.

cevich:
> You're probably going to want to replace -t ...with...

Also good advise.  Though lots of other things use that podman build Makefile recipe, and expect regular images.  Hrmmm, I guess lets see when you get that far.

nalind:
If there's no option that uses an RPM package that's part of the distribution, running
 `sudo podman run --net=host --rm --privileged multiarch/qemu-user-static -p yes`
or some equivalent should work approximately as well, if SELinux isn't going to prevent its binaries from interacting with binaries in other containers.

lsm5:
@cevich shouldn't the new commit now run everything in `centos-stream-9` ? Re-running `Build fedora_podman container` with Terminal still shows me 
```
# cat /etc/centos-release 
CentOS Stream release 8
```

cevich:
> shouldn't the new commit now run everything

Not everything, because the family name is used: https://github.com/containers/automation_images/blob/main/.cirrus.yml#L79

It's possible somebody else was running an image build in another PR, and so the "latest" image-builder VM wound up being theirs not yours.  However, use of the family name is necessary to support the `[CI:TOOLING]` feature :confounded:

Hmmm.  I'd suggest either making the update to C9s a separate PR, or using nalin's `multiarch/qemu-user-static` container suggestion.  I've done that before in different contexts, it works well (though it's confusing to understand *how* it works).  If disabling SELinux on the image-builder is needed, I'm fine with that.

cevich:
Closing as "not being worked on".  Please re-open if that changes.

PR #160, opened 2022-08-01
[CI:TOOLING] Add support for EC2 instances in get_ci_vm

cevich:
Add necessary updates to support creation/use of EC2 CI VMs with
containers-org `hack/get_ci_vm.sh` tool.

Notes:
* Only works for repositories that have Cirrus-CI setup to run tasks in
  EC2
* Any `.cirrus.yml` tasks running in EC2 must define a `$EC2_INST_TYPE`
  env. var (in the YAML)
* Also requires a minor update to the `hack/get_ci_vm.sh` script
  contents.

Other changes:
* Moved addition of `entrypoint.sh` script in Containerfile to speed up
  build/test development cycles.
* Added `awscli` tool and necessary dependencies to containers `setup.sh`.
* Updated a few GCP lines and messages to improve consistency with EC2.
* Renamed/Updated `create_gcevm()` to a generic `create_vm()` that works
  with both clouds.
* Updated `fini()` and `fini_vm()` to be generic and work with both
  clouds.
* Added some very basic EC2 unit-tests.  More would be nice, but the
  EC2 interactions are far more complex to mock than GCE.

### DIFF
--- a/get_ci_vm/entrypoint.sh
+++ b/get_ci_vm/entrypoint.sh
@@ -30,6 +30,12 @@ READING_DELAY="${READING_DELAY:-5s}"
 # Set non-zero to enable debugging
 A_DEBUG="${A_DEBUG:-0}"
 
+# Required to be non-empty, updated by the supports_apiv*() functions
+GET_CI_VM="${GET_CI_VM:-0}"
+
+# Used to cache status of local setup check
+CREDENTIALS_ARE_VALID=0
+
 # apiv1 expects to receive the following env. vars. when called with --config
 # and $GET_CI_VM is set to 1
 #
@@ -42,7 +48,7 @@ UPSTREAM_REPO=""
 #
 # Optional:
 #
-# path where setup script persists it's env. vars
+# path where setup script persists its env. vars
 CI_ENVFILE=""
 #
 # 'gcevm' instances type required:
@@ -62,6 +68,17 @@ GCLOUD_PROJECT=""
 # GCE Project ID where images are stored
 GCLOUD_IMGPROJECT=""
 
+# apiv2 expects to receive the following additional env. vars. when
+# called with --config and $GET_CI_VM is set to 2
+#
+# Required:
+#
+# This holds the name of the AWS config/credentials profile.  This
+# is needed because the CLI keeps this data all in a single pair of
+# files and the calling user may have AWS already setup for other
+# purposes.
+AWS_PROFILE=""
+
 status() {
     local message="$1"
     local red="\e[1;31m"
@@ -74,17 +91,43 @@ status() {
 }
 
 # Each repository behaves slightly differently, confirm
-# compatibility before loading repo-specific settings
+# compatibility before loading repo-specific settings.
+# Specifically, this requires the hack script to define
+# a bunch of variables and support specific command-line
+# options.  Please see an existing/working example for
+# an idea of what the requirements are.
 supports_apiv1() {
+    req_env_vars REPO_HACK_SCRIPT GET_CI_VM
     dbg "Examining $REPO_HACK_SCRIPT for apiv1 support"
-    req_env_vars REPO_HACK_SCRIPT
     if [[ ! -r "$REPO_HACK_SCRIPT" ]]; then
         die "Can't find '$REPO_HACK_SCRIPT' in source repository."
+    elif ((GET_CI_VM)); then  # Already verified & set
+        return 0
     elif grep -Eq '^# get_ci_vm APIv1' "$REPO_HACK_SCRIPT"; then
         dbg "Detected apiv1 support."
+        export GET_CI_VM=1  # checked by hack/get_ci_vm.sh
         return 0
     fi
     dbg "Did NOT detect apiv1 support."
+    return 1
+}
+
+# Requires apiv1 support + specific EC2 support changes
+# in the repository.  Specifically, the hack script
+# needs to define an AWS_PROFILE and the .cirrus.yml
+# EC2 task(s) need to define EC2_INST_TYPE.
+supports_apiv2() {
+    req_env_vars REPO_HACK_SCRIPT GET_CI_VM
+    if ! ((GET_CI_VM)); then
+        die "$REPO_HACK_SCRIPT does not support apiv1"
+    fi
+    dbg "Examining $REPO_HACK_SCRIPT for apiv2 support"
+    if grep -Eq '# get_ci_vm APIv2' "$REPO_HACK_SCRIPT"; then
+        dbg "Detected apiv2 support."
+        export GET_CI_VM=2  # checked by hack/get_ci_vm.sh
+        return 0
+    fi
+    dbg "Did NOT detect apiv2 support."
     return 1
 }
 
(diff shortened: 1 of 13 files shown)


cevich:
***HOW TO TEST:*** See https://github.com/containers/podman/pull/15145#issuecomment-1203183272

cevich:
/sigh/ yes, it really does say **668** new lines :disappointed: It's very likely far to complex for anybody to give more than a cursory look through.  It took me some serious hair-pulling over several fully-committed days to get this working as-is.  I tried to comment and document as best as I could.  The new unit-tests are definitely also lacking, but not out of laziness.  I simply found I would need to mock-out so much of the new code, automated tests would have little value :disappointed_relieved:

lsm5:
> _**HOW TO TEST:**_
> 
> 1. Run `make get_ci_vm IMG_SFX=latest` (you'll need podman installed locally)
> 2. Tag the resulting image `quay.io/libpod/get_ci_vm:latest`

This step is handled in the first step itself. Less work for the tester :)

> 4. Run `hack/get_ci_vm.sh Build for fedora-36-aarch64` (you'll need an AWS access key and secret).
I see this towards the end (ami id replaced with FOO):

```
# Configuring shallow clone of local repository
Switched to a new branch 'get_ci_vm'

# Configuring/setting up freshly created lsm5-ami-FOO
+ Transferring repo. tarball to instance
scp: unrecognized option: k
usage: scp [-346ABCOpqRrsTv] [-c cipher] [-D sftp_server_path] [-F ssh_config]
           [-i identity_file] [-J destination] [-l limit]
           [-o ssh_option] [-P port] [-S program] source ... target

# Finalizing get_ci_vm
Offering to delete lsm5-ami-FOO (might take a minute)
Note: It's perfectly safe to answer 'N'.  Simply re-run script again
with the same task argument, and you can access and/or delete the VM.

Are you sure you want to remove lsm5-ami-FOO (y/N):
```

lsm5:
Also, can the script be made to use my `~/.aws/credentials` info before prompting me to enter the key id and secret?

get_ci_vm/entrypoint.sh
```diff
@@ -76,18 +90,37 @@ status() {
...
         dbg "Detected apiv1 support."
+        export GET_CI_VM=1  # checked by hack/get_ci_vm.sh
         return 0
     fi
     dbg "Did NOT detect apiv1 support."
     return 1
 }
 
+# Assumes apiv1 support
+supports_apiv2() {
```
edsantiago, line 109: At first glance the comment looks like a typo (1 vs 2); after reading the function, the comment makes no sense at all. My best guess is, "No need to check for apiv1, it is assumed that we always support that", but if that's what the comment means, there isn't really a need for it I think. Or I might be wrong. Could you please rewrite it, keeping in mind that the reader may not know what "apiv1" and "v2" mean?

edsantiago (commented):
Submitting interim review feedback because of my terror that you'll push something and github will lose all my feedback so far.

get_ci_vm/entrypoint.sh
```diff
@@ -366,82 +525,352 @@ Note: If asked to set a SSH-key passphrase, DO NOT SET ONE, it
...
+    # It's possible that set_ec2_inst_id() looked up a transitioning or temporarily
+    # useless ID.  Force it to do anotherlookup the next time it's called.
+    EC2_INST_ID=""
+    return 1
+}
+
+# GCP provides a handy wrapper for ssh and scp, for AWS it's all DIY.
+# This function initializes both $SSH_CMD and $SCP_CMD on success,
+# when the initial reachability/state of the VM after creation is
+# unknown for some time period.  Accepts zero or more arguments,
```
cevich, line 688: fix grammar

get_ci_vm/entrypoint.sh
```diff
@@ -366,82 +525,352 @@ Note: If asked to set a SSH-key passphrase, DO NOT SET ONE, it
...
+    # other state was caused by user-actions, so they're responsible for any
+    # odd script behavior (i.e. re-running script "too soon" after rebooting a VM)
+    if set_ec2_dns_name; then
+        warn "Reusing pre-existing instance '$INST_NAME' with ID '$EC2_INST_ID' at '$DNS_NAME'.  If it's inaccessible, get_ci_vm may behave oddly."
+        return 1  # create_vm() expects this behavior
+    fi
+
+    # For whatever reason the documented "shorthand" syntax doesn't work
+    # reliably with even slightly complex options.  Most of this is basic
+    # but must be fully spelled out in all it's glorious JSON detail.  If
```
cevich, line 725: s/it's/its/

get_ci_vm/entrypoint.sh
```diff
@@ -176,6 +217,114 @@ has_valid_credentials() {
...
+        dbg "AWS config/credentials are missing"
+        return 1
+    elif [[ ! -r "$EC2_SSH_KEY" ]] || [[ ! -r "${EC2_SSH_KEY}.pub" ]]; then
+        dbg "EC2 ssh keys are missing"
+        return 1
+    else
+        # It's possible the config. is incomplete, verify minimums
+        for _cfgvar in aws_access_key_id aws_secret_access_key region; do
+            # Detect both errors on missing profile and/or missing files.
+            if ! $AWSCLI configure get $_cfgvar &> /dev/null; then
```
edsantiago, line 236: Is the difference between "missing" and "empty" important? I will assume so, but please remember that it may be important to you but not to anyone else, and when an end user sees this error, they will report it as "something about var", and they won't remember whether the magic word was "missing" or "empty", so it might be nice to provide a more verbose explanation such as "AWS variable '$_cfgvar' is not present in the configuration" and "AWS variable '$_cfgvar' is present in the configuration, but empty", along with perhaps a "WEIRD" or something indicating how likely/impossible this is to happen, and what to do about it.

Oh, also the duplicate call is unnecessary:
```bash
if ! foo=$(AWS...); then a; elif [[ -z "$foo" ]]; then b ...
```

cevich:
> This step is handled in the first step itself. Less work for the tester :)

Sometimes it did that for me, other times not, I didn't investigate why.

> I see this towards the end (ami id replaced with FOO):
> 
> scp: unrecognized option: k
> usage: scp [-346ABCOpqRrsTv] [-c cipher] [-D sftp_server_path] [-F ssh_config]
>            [-i identity_file] [-J destination] [-l limit]
>            [-o ssh_option] [-P port] [-S program] source ... target
> 

Hmmm, I didn't remember seeing that....oh I see the problem, okay it will be fixed in my next push.

> EDIT: hitting either of y/N makes no visible difference except y telling me Deleting lsm5-ami-FOO and returning to prompt, while N silently returns to prompt.

okay, added a message to say "Keeping VM as-is, please re-run script later to remove it when you're done."  Will be in my next push.

cevich:
> Submitting interim review feedback because of my terror that you'll push something and github will lose all my feedback so far.

Oh, sure I can avoid pushing anything, please continue.  I can simply queue up fixes locally and wait.  Thanks for taking a look :smiley:

cevich:
> Also, can the script be made to use my ~/.aws/credentials info before prompting me to enter the key id and secret?

I thought quite a bit about this and decided it was better to err on the side of caution.  Doing it this way also matches how we handle GCP configuration (segregated by profiles).

get_ci_vm/entrypoint.sh
cevich: Oh, no this comment is wrong.  I'll fix.

get_ci_vm/entrypoint.sh
```diff
@@ -106,17 +139,25 @@ init() {
...
+        die "Repository hack/get_ci_vm.sh not compatible with Api v1"
+    fi
+
+    if supports_apiv2; then  # sets/exports GET_CI_VM=2
+        "$REPO_HACK_SCRIPT" --config > $_TMPDIR/apiv2.sh
+        dbg "Loading apiv2 vars:
+$(cat $_TMPDIR/apiv2.sh)"
+        # shellcheck disable=SC1090
+        source $_TMPDIR/apiv2.sh
+        req_env_vars AWS_PROFILE
```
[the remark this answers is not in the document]
cevich: Yes both are needed (I fixed the `supports_apiv2()` comment to reflect this)

get_ci_vm/entrypoint.sh
cevich: ...Updated function to treat both unset and empty the same (`return 1`).

cevich:
@edsantiago I need to force-push fixes for your suggestions, and to a problem lokesh and charlie hit when trying to use it.  Old commit was 180c0b9

cevich:
@edsantiago we good to merge this then?

edsantiago:
I'm still reviewing (the ssh keygen took me ages) but am getting too many interruptions. Sure, merge it, I guess if there are problems they can be fixed later. Nice work, but sheesh, my brain hurts.

/lgtm

cevich:
> Nice work, but sheesh, my brain hurts.

Yeah it's bad.  And I've got more work upcoming that involves the `aws` CLI.  Very much not looking forward to it.  Anyway, thanks for all the help reviewing, I know it is/was a lot lot.

PR #162, opened 2022-08-02
[CI:DOCS] GHA: Re-use identical workflow from buildah repo

cevich:
It's a PITA to maintain duplicate code across repos.  Relatively
github-actions added a feature that allows re-using workflows
from other repos.  Use that here to reduce duplication.


cevich:
Manually tested and confirmed functional: https://github.com/containers/automation_images/actions/runs/2784986707

PR #163, opened 2022-08-04
Set EC2 cirrus-agent SELinux context to unconfined

cevich:
Long ago, many test failures and other problems were experienced in
Cirrus-CI managed google-cloud VMs.  None of the problems were
reproducible manually.  It was discovered that because the cirrus agent
starts from a metadata-downloaded script, it was executing with a more
restrictive SELinux type.  This is not the case when running tests
manually, where root sshs in.

It's been observed recently, a similar situation may be occurring in
EC2.  However, in this case, the agent is started by cloud-init, and was
observed operating with the type `cloud_init_t`.  In case this is the
source of trouble now or in the future, fix the setup to match GCP.

### DIFF
--- a/base_images/fedora_base-setup.sh
+++ b/base_images/fedora_base-setup.sh
@@ -32,6 +32,7 @@ else  # A VM
     # shellcheck disable=SC2154
     if [[ "$PACKER_BUILD_NAME" =~ "aws" ]]; then
         echo "WARN: AWS EC2 Instance Connect not supported on Fedora, use cloud-init."
+        PKGS+=(policycoreutils-python-utils policycoreutils)
     else  # GCP image
         PKGS+=(google-compute-engine-oslogin)
         if ((OS_RELEASE_VER<35)); then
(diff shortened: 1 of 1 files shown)


cevich:
Ref: https://github.com/containers/podman/pull/15179

cevich:
@cdoern okay, these AWS images (assuming the build works) should be running the cirrus-agent unconfined.  You can confirm this in your PR (after swapping in these images) using the "rerun in terminal" trick.  Simply so a `ps -wauxZ | grep agent` and confirm it's `unconfined_t`.

cevich:
force-push: Fixed woopsie.

cevich:
@cdoern finally it finished.  The bits you need to stuff into `.cirrus.yml` should be just:

`c4505025664778240`

and

|Name|ID|
|---|---|
|**fedora-aws**|`ami-0be69d43075633302`|
|**fedora-podman-aws-arm64**|`ami-099866e4c549960af`|

There should be comments making it fairly clear where the IDs go.  Commit and push that in https://github.com/containers/podman/pull/15179 and maybe(?) the results will improve?

If not...well...we know for sure it's not an SELinux related problem.  Also, we probably should have this change in our AWS images anyway, so no biggie.

cdoern:
@cevich in the yml, where do the ami's go? is it `FEDORA_AMI_ID` and `FEDORA_AARCH64_AMI_ID`?

cevich:
Closing: These EC2 images are broken, the `cloud-final.service` fails to start - causing Cirrus to "hang" (no agent).  I tried relabeling the binary `unconfined_t` but `chcon` says: permission denied.

cevich:
Force-push:  Found a fix that seems to allow cloud-init services to start:  Added a custom default `bin_t` context on `/usr/bin/cloud-init` followed by a `restorecon`.

cevich:
force-push: :roll_eyes: forgot a package.

cdoern:
I'll try these new images see if they do anything

cevich:
Thanks @cdoern you read my mind.

>  see if they do anything

Hopefully they will at least boot this time :grinning:

cevich:
I'm going to merge this, despite it not fixing https://github.com/containers/podman/issues/15074 since passed all other podman tests and it improves runtime consistency between google and AWS.

PR #166, opened 2022-08-10
See if Ubuntu has new runc

edsantiago:
This is an empty commit and empty PR. Purpose is to see if Ubuntu
has picked up a fixed runc, which would allow us to close

   https://github.com/containers/podman/issues/15025


edsantiago:
Broken runc version is `1.1.0-0ubuntu1-amd64`. If new build gets same version, close this PR.

edsantiago:
In [this line](https://cirrus-ci.com/task/4679992683528192?logs=main#L978) in the `ubuntu Cache Image` job I see:
```
    ubuntu: Unpacking runc (1.1.0-0ubuntu1) ...
```
...which suggests that runc is still broken. But I'm not sure I got the right one (the `Base` one does not show `runc` anywhere in its logs). @cevich if you agree with my assessment, please just close this PR.

cevich:
It's normal not to see this in the base image.  For the future, a (potentially) faster way to learn this is:

Fire up the container-analog to the VM you're interested in.  For example, `quay.io/libpod/fedora_podman:c<IMG_SFX>`, or `prior-fedora_podman`, or `ubuntu_podman`.  You can just run `apt-get update` directly and see what happens.

Just be aware the container images are fairly large, like 600-800mb.  If you're on a slower network, `hack/get_ci_vm.sh` is your friend.  Along with the same 'apt-get update` of course.

I would just close this, but, do you have a sense of how much coverage we're loosing?  

I fully anticipate it could be months (or never) that runc gets updated in Ubuntu.  That's been the pattern in the past, and why we've needed Lokesh to custom build one :disappointed:

cevich:
"How-much" clarification:  For example, is it just one or two tests that are otherwise executed elsewhere (like in Fedora) under crun?

edsantiago:
Seven tests are currently skipped on cgroupsv1:
```
030-run.bats-@test "podman run - uidmapping has no /sys/kernel mounts" {
030-run.bats-@test "podman run update /etc/hosts" {
170-run-userns.bats-@test "podman --group-add without keep-groups while in a userns" {
170-run-userns.bats-@test "rootful pod with custom ID mapping" {
400-unprivileged-access.bats-@test "podman container storage is not accessible by unprivileged users" {
500-networking.bats-@test "podman networking: port with --userns=keep-id for rootless or --uidmap=* for rootful" {
500-networking.bats-@test "podman run /etc/* permissions" {
```
Closing, since the issue is not resolved.

PR #169, opened 2022-08-16
Support optional EC2 images in imgts container

cevich:
AWS EC2 has a separate mechanism for recording an AMI's last-used
timestamp.  However, other important metadata is also maintained on
images such as `permanent=true` (or not).  For repos. which utilize EC2,
add support for the imgts container updating EC2 images metadata.

Also, add a task to exercise the imgts container in a [CI:TOOLING]
context (where the image is built).  Set this up as a dry-run
so it doesn't depend on the images actually existing.

### DIFF
--- a/imgts/entrypoint.sh
+++ b/imgts/entrypoint.sh
@@ -3,19 +3,19 @@
 # This script is set as, and intended to run as the `imgts` container's
 # entrypoint. It's purpose is to operate on a list of VM Images, adding
 # metadata to each.  It must be executed alongside any repository's
-# automation, which produces or uses GCP VM Images.
+# automation, which produces or uses GCP VMs and/or AWS EC2 instances.
 #
 # N/B: Timestamp updating is not required for AWS EC2 images as they
 # have a 'LastLaunchedTime' attribute which is updated automatically.
+# However, updating their permanent=true tag (when appropriate) and
+# a reference to the build ID and repo name are all useful.
 
 set -e
 
-# shellcheck source=./lib_entrypoint.sh
+# shellcheck source=imgts/lib_entrypoint.sh
 source /usr/local/bin/lib_entrypoint.sh
 
 req_env_vars GCPJSON GCPNAME GCPPROJECT IMGNAMES BUILDID REPOREF
-
-gcloud_init
 
 # Set this to 1 for testing
 DRY_RUN="${DRY_RUN:-0}"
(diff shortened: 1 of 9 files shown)


cevich:
@edsantiago this one needs one more round of manual testing, but more importantly could use some extra diff-scrutiny.  This image is used by every branch of every cirrus-ci enabled repo.  I intend to merge it (thereby deploying the image) only after letting the entire team to BOLO related breakage (it will be basically instant).  In any case, I have another PR to work on in the meanwhile, so there's not a major rush to get this in (though my other PR does depend on this one).

cevich:
Force-push: Minor rebase on main, one-line conflict fixed.

imgts/entrypoint.sh
```diff
@@ -107,32 +107,81 @@ if is_release_branch_image $BUILDID; then
     SET_PERM=1
 fi
 
-if ((DRY_RUN)); then GCLOUD='echo'; fi
+if ((DRY_RUN)); then
+    GCLOUD='echo'
+    AWS='echo'
```
edsantiago, line 112: It doesn't look like any of the `$GCLOUD`/`$AWS` incantations are logged, but in case some day they are, a future maintainer might appreciate something like
```
    GCLOUD="echo $GCLOUD"
    AWS="echo $AWS"
```
Otherwise, the only thing logged is the arguments. It might be possible for the maintainer to figure out "is this an aws or gcloud command" based on the context, and perhaps the args also, but why not be kind to them in anticipation?

imgts/entrypoint.sh
```diff
@@ -107,32 +107,81 @@ if is_release_branch_image $BUILDID; then
...
+# Not all repos use EC2 instances, only touch AWS if both
+# EC2IMGNAMES and AWSINI are set.
+if [[ -n "$EC2IMGNAMES" ]]; then
+    msg "---"
+    req_env_vars AWSINI BUILDID REPOREF
+
+    if ! ((DRY_RUN)); then
+        aws_init
+        # aws_init() has no output because that would break in other contexts.
+        msg "Activted AWS CLI for service acount."
```
edsantiago, line 152: typo, `Activ>>a<<ted`

edsantiago (commented):
Thanks for your patience; this is always really hard for me to review. A few comments, no showstoppers. I have no way to test this, and no knowledge of aws/gcloud, so my review is limited to the very basics.

imgts/entrypoint.sh
```diff
@@ -107,32 +107,81 @@ if is_release_branch_image $BUILDID; then
...
+            amiid="dry-run-$image"
+        elif ! amiid=$(get_ec2_ami "$image"); then
+            ERRIMGS+=" $image"
+            continue
+        fi
+
+        # AWS deliberately left unquoted for intentional word-splitting
+        # shellcheck disable=SC2206
+        awscmd=(\
+            $AWS ec2 create-tags "$dry_run"
```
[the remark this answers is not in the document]
cevich: Ohh good catch, it was set in a prior iteration I dropped it in favor of the simpler `AWS='echo'` approach.  I'll fix it.

imgts/entrypoint.sh
cevich: Sure thing, it's a good suggestion.

cevich:
> I have no way to test this, and no knowledge of aws/gcloud, so my review is limited to the very basics.

There isn't really an easy way to test it manually, other than carefully setting all the env. vars.  Which is why I added the "test_imgts_task" in `.cirrus.yml`.  There's a separate task that does an actual run when VM images are built, so I can trigger that to verify the correct labels are set in both GCP and EC2...

cevich:
...force-pushed fixes + setup this PR to do a full image-build.  That will allow manual validation the built images got the right labels/tags assigned.

cevich:
@edsantiago I think this one is good to go:
* [Test image timestamp/metadata updates](https://cirrus-ci.com/task/4898708121714688) passed with expected output
* [Update image timestamp/metadata](https://cirrus-ci.com/task/6024608028557312) passed with expected output.  I also examined a few of the images in the cloud and everything looks good.

edsantiago:
LGTM

PR #171, opened 2022-08-18
[CI:DOCS] Fix error caused by multi-line $AWSINI contents

cevich:
This is somehow causing a silent failure of the workflow, with an error
coming from podman:

```
Error: error parsing env file "/tmp/env_file_<BLAH>": name "output "
has white spaces, poorly formatted name
```

This is due to the $AWSINI contents being mult-line.  Fix this by
making github-actions set the secret env. vars.  Then after making
sure they're exported, let podman figure out how to handle their
contents.

Also, split the data collection from data processing tasks to improve
debug-ability.  Rename the output file and artifact to indicate their
expected contents better.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #172, opened 2022-08-18
[CI:DOCS] Fix passing secret env. vars. into podman

cevich:
Fix issue introduced in https://github.com/containers/automation_images/pull/171

At the time of this commit, the default github-action runners are using
an ancient version of podman.  This version does not support using the
`-e ENVAR` feature to slurp up env. var values from the environment.
They must be set on the command-line.


cevich:
No way to test this other than from the 'main' branch :angry:

cevich:
Manually triggered: https://github.com/containers/automation_images/actions/runs/2883934131

Issue #173, opened 2022-08-18
Orphanvms github-action workflow is broken


cevich:
This started by my noticing a silent-failure in the jobs, caused by a multi-line value not making it through the `--env-file` podman option.  An attempt at a fix was made: https://github.com/containers/automation_images/pull/171

Unfortunately the fix also failed, due to the "ubuntu-latest" runner providing podman 3.4.4 which apparently doesn't support slurping env. vars. up from the host environment (`-e NAME` argument).  An attempt was made to revert to simply passing values in on the command line (`-e NAME=VALUE` form):  https://github.com/containers/automation_images/pull/172

Sadly this attempt also failed, due to another form of the original problem, passing the multi-line `AWSINI` value on the command-line.  Now I'm at my whits end.  It's incredibly frustrating that the only way to test if a fix works, is to merge a PR :angry:

cevich:
OMG.  According to https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources `ubuntu-22.04` is available, but `ubuntu-latest` still points at `20.04`.  WTF GITHUB!?!?!?!?!

cevich:
Double OMG: Apparently the Ubuntu 22.04 runner [also only has podman 3.4.4](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md).

PR #174, opened 2022-08-18
[CI:DOCS] Use --env-host for podman 3.4.4

cevich:
An attempt to re-fix https://github.com/containers/automation_images/pull/171
Ref: https://github.com/containers/automation_images/issues/173


cevich:
Manually triggered https://github.com/containers/automation_images/actions/runs/2884141472

PR #175, opened 2022-08-18
[CI:DOCS] Fix missing GCPPROJECT requirement in orphanvms

cevich:
This var. was always required by gcloud-init() but by re-initializing
inside a loop of other projects, gcloud apparently ignored it being an
incorrect value.  Clarify the situation by making the env. var. required
and changing the loop variable name to avoid any confusion.

This should hopefully/finally fix https://github.com/containers/automation_images/issues/173


cevich:
Manually triggered run: https://github.com/containers/automation_images/actions/runs/2884241348

Issue #135, opened 2022-05-26
Re-use `podman/*_release` contents in packer


cevich:
There's no practical reason why this can't be achieved.  It would reduce the mental-drain when updating from one release to another - as in, you'd only need to update the version in one place instead of three.  Probably the safest/easiest way is to consume the contents in the Makefile, and pass them as vars into packer.

PR #178, opened 2022-08-23
Use base_image/cloud.yml as release version SoT

cevich:
Fixes: #135

Previously, a maintainer integrating a new OS release was required to
update multiple files.  Reduce this burden by encoding the
release-version in base-image YAML and retrieving it as needed.  This
also has the added side-benefit of allowing the release version to be
tagged on the images themselves for reference.


lsm5 (approved):
LGTM after happy tests. This is really convenient!!

cevich:
> after happy tests

Right...that's almost always the actual hard part :rofl:

cevich:
Force-pushed:  More `Makefile` updates - I missed a few more `base_images/cloud.yml` dependencies.

cache_images/cloud.yml
```diff
@@ -30,9 +35,10 @@ builders:
       zone: 'us-central1-a'
       disk_size: 20  # REQUIRED: Runtime allocation > this value
       disable_default_service_account: true
-      labels:  # For the VM
+      labels: &gce_labels # For the VM
         sfx: '{{user `IMG_SFX`}}'
         src: '{{ build_name }}-b{{user `IMG_SFX` }}'
+        release: 'ubuntu-{{user `FEDORA_RELEASE`}}'
```
cevich, line 41: Woops, I'll fix that in the next commit.

cevich:
@lsm5 okay this seems to be working, but I'm not completely happy about it and could use your input.

The main thing is (realized in hind-sight):  We don't actually use any of the data from `image_info.yml` anywhere except from within the `Makefile`.  So I could simplify things quite a bit by just moving it all there, dropping the `get_image_info.sh` script, and simply require maintainers to update the `Makefile` any time changes are needed.

OTOH, the only real advantage I can see of keeping the details in a separate, simple YAML file is to reduce eye-strain.  In other words, avoid exposing casual-maintainers to the complexity of the `Makefile`.  This seems like a fairly weak argument to me though.

WDYT?

cevich:
@edsantiago I'd appreciate your opinion ([see above comment](https://github.com/containers/automation_images/pull/178#issuecomment-1227830262)) as well (no need for a code review).  Should I just simplify all the image and release info details directly into the `Makefile`, or do you think it's helpful to have it separate (or something else I haven't thought of).

Point is (as the PR description says): Make updating OS images (e.g. for a new version of Fedora) as simple as possible by keeping them in one place instead of spread among many.

Makefile
```diff
@@ -74,6 +77,22 @@ override _HLPFMT = "%-20s %s\n"
...
+override _fedora_image_url = $(call get_image_info,fedora.image_url)
+override _fedora_csum_url = $(call get_image_info,fedora.csum_url)
+override _fedora_ami = $(call get_image_info,fedora.ami)
+override _fedora_arm64_ami = $(call get_image_info,fedora.arm64_ami)
+override _prior_fedora_release = $(call get_image_info,prior_fedora.release)
+override _prior_fedora_image_url = $(call get_image_info,prior_fedora.image_url)
+override _prior_fedora_csum_url = $(call get_image_info,prior_fedora.csum_url)
+override _ubuntu_release = $(call get_image_info,ubuntu.release)
+override _ubuntu_base_family = $(call get_image_info,ubuntu.family)
+
```
edsantiago, line 95: Just from a cursory first-pass review, I find this (plus the `get` script) kind of fragile and hard to read/maintain. My vote is to ditch the separate yaml and script.

cevich: Ya that's what I was thinking too. Okay I'll simplify that into the Makefile then.

cevich:
Force-push: Simplify, define build variables in the `Makefile`.  Also, added a definition for the CentOS-stream release (used in a few containers and the `image_builder` image).

.gitignore
```diff
@@ -1,2 +1,3 @@
 */*.json
+image_info.json
```
edsantiago, line 2: You probably don't need or want this any more. (And wasn't it yaml anyway, not json?)

base_images/cloud.yml
```diff
@@ -17,31 +17,19 @@ variables:  # Empty value means it must be passed in on command-line
     # Allows providing handy cross-reference to the build log
     CIRRUS_TASK_ID: "{{env `CIRRUS_TASK_ID`}}"
 
-    # BIG-FAT-WARNING:  When updating the image names and/or URLs below,
-    # ensure the distro version numbers contained in the `podman/*_release`
-    # files exactly match.  These represent the container base-image tags
-    # to build from - just as the sources below are the base-images to
-    # start from building VM images.
+    # See image_info.yml
```
edsantiago, line 20: Obsolete comment?

cache_images/cloud.yml
```diff
@@ -16,6 +16,11 @@ variables:  # Empty value means it must be passed in on command-line
     # Allows providing handy cross-reference to the build log
     CIRRUS_TASK_ID: "{{env `CIRRUS_TASK_ID`}}"
 
+    # See image_info.yml
```
edsantiago, line 19: obsolete comment?

edsantiago (commented):
Way over my head. I don't understand the `AWS_SHARED_CREDENTIALS_FILE` stuff, nor why URLs changed from `/build` to `/task`, ... but from what little I can understand, I have a few questions

.gitignore
cevich: it was both, but good catch, I'll remove it.

base_images/cloud.yml
cevich: Yep, thanks.

cevich:
> why URLs changed from /build to /task

Oh, this is to assist maintainers when viewing images on AWS or GCE, they have an immediate link right to the build-logs (task) for the image, rather than the (less immediately useful) build (collection of tasks).  Also note, it's similar to the recently updated table of image names and IDs (comment-bot) - it now contains links to the cirrus-task logs for each image in addition to the (less useful) build.

> I don't understand the AWS_SHARED_CREDENTIALS_FILE stuff

The `AWS_SHARED_CREDENTIALS_FILE: notused` addition is needed because [the `packer_build` function checks the value is non-empty ](https://github.com/containers/automation_images/pull/178/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R190)- which is useful for every image build except the `image_builder` target (and image).

cevich:
This seems to be working well, merging.

PR #177, opened 2022-08-22
Update package dependencies

lsm5:
Install python3-devel and do not download podman-* packages as they can
be fetched from dnf.

oci-umount should no longer be needed.

Fixes: #159 and #161

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

PR #179, opened 2022-08-23
Enable Fedora 37 images

lsm5:


lsm5:
Looks like packit's recommended testing might need quite some rewriting of tests (at first glance), so I'm gonna try to proceed with both approaches in parallel. Scratch build testing via copr is pretty simple so we can get that part going already without too much fuss.

lsm5:
piggybacks on https://github.com/containers/automation_images/pull/177

cevich:
Apologies if I wasn't clear:  The problem here is with the EC2 images.  I opened a ticket in the Fedora cloud-sig to have the F37 beta images imported.  But based on previous requests, this probably won't happen for weeks/months :cry: 

At the same time everywhere in CI, there are lots of "coincidental" (for lack of a better term) associations with image contents.  For example, in podman CI the `$FEDORA_NAME` and `$DISTRO_NV` variables are assumed to be reflective of what's actually inside the images.  Sometimes this is verified, sometimes not.

Now, before you say "We can just break this down, and have the EC2 images lag behind"...On one occasion I remember too well, these associations were broken on purpose (due to an update problem).  CI worked fine at first for a week or so, then a months-long nightmare unfolded because the "disconnected" images weren't being tested.  As soon as I tried to update them, TONS of tests broke :angry: 

Fixing this is going to be complex (based on my research last week).  Because our AWS images are multi-arch, we cannot rely on the x86_64 QEMU builder to bring them in like we do for Fedora in GCE.  I think we need to add a new target, a "PRE-base_images" stage, where [packer imports the generic cloud images into EC2](https://www.packer.io/plugins/post-processors/amazon).  As part of the import, the AMI can be tagged for pickup in `base_images/cloud.yml` using a `source_ami_filter` block (like we do in `cache_images/cloud.yml`).

Whew!  Clear as mud?  Have a better/simpler idea?

cevich:
Here's what I mean, in YAML form: https://paste.centos.org/view/45c5d996
(I have ZERO idea if that will actually work, and the account packer uses will need an IAM profile added to permit S3 access)

cevich:
Another idea I had this morning: What if for F37 and rawhide even, we used the currently provided Fedora images and upgraded them in packer?  That would avoid the messiness of an import at the cost of a reboot (which packer is happy to allow) and some more scripting.  About the only down-side is my (maybe false) perception the result will be less "pure" than an import from upstream cloud images.  Maybe this isn't important for our purposes though.

cevich:
@lsm5 Alight, well since I had handled all that import yaml already, I decided to try doing it the hard way: #182 (fingers-crossed)

Note: This could serve as a basis for supporting rawhide as well, since they do produce cloud images that could be similarly imported: https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Cloud/

lsm5:
i might take a while to process all this :smile: or maybe I'll have to actually see failures to realize what you mean.

cevich:
>  actually see failures to realize what you mean.

Yeah, sorry, it's kind of tricky.  Basically it all boils down to: We don't want our images to get out-of-sync with each other.  So "Fedora 37" means exactly the same thing (in terms of image-contents, on GCP as on AWS as in a container).  Actual code-bugs in a PR are hard enough for developers to debug w/o having to also worry about: "Is the OS environment in this task, the same as this other task which passed?".

So what I'd like to do (assuming I can get #182 to work) is simply import the images into AWS and GCE from the same set of URLs.  That way everything is same-same and there's no way to question where any image came from.  It also means we don't need to wait on the fedora cloud-sig to make beta AMI's available.

I _think_ this approach will be slightly cleaner than my "upgrade" idea.  But we can keep that as a fallback option.

cevich:
Closing in favor of https://github.com/containers/automation_images/pull/182

Issue #186, opened 2022-08-26
Obsolete image detection throws silent errors


cevich:
Example log: https://cirrus-ci.com/task/5398202080821248

```
...cut...
Obsoleting EC2 image ami-0390056032f925092:
 Used over 30 days ago on 2022-06-13; Tags: permanent=,build-id=,repo-ref=,automation=true,lastLaunchedTime=2022-06-13
{
    "Return": true
}
Obsoleting EC2 image ami-0266d0f7e46ecd601:
 Used over 30 days ago on 2022-06-14; Tags: permanent=,build-id=,repo-ref=,automation=true,lastLaunchedTime=2022-06-14
{
    "Return": true
}
Obsoleting EC2 image fedora-netavark-aws-c5958996065517568:
 Missing 'lastLaunchedTime' metadata; Tags: permanent=,build-id=,repo-ref=,automation=true
An error occurred (UnauthorizedOperation) when calling the EnableImageDeprecation operation: You are not authorized to perform this operation. Encoded authorization failure message: ocs...xTg
```

It seems the removal loop should be counting aws CLI output that doesn't say `"Return": true`, then throwing an actual error at the end (so all removals have a chance of happening).

cevich:
Hint: The encoded permission errors likely have to do with those images missing certain required tags.  Decoding one of them didn't give any obvious clues.  At the same time, they also provide pretty limited information:

```json
{
  "allowed": false,
  "explicitDeny": false,
  "matchedStatements": {
    "items": []
  },
  "failures": {
    "items": []
  },
  "context": {
    "principal": {
      ...cut...
    },
    "action": "ec2:EnableImageDeprecation",
    "resource": "arn:aws:ec2:us-east-1::image/*",
    "conditions": {
      "items": [
        {
          "key": "aws:Region",
          "values": {
            "items": [
              {
                "value": "us-east-1"
              }
            ]
          }
        },
        {
          "key": "aws:Service",
          "values": {
            "items": [
              {
                "value": "ec2"
              }
            ]
          }
        },
        {
          "key": "ec2:ImageID",
          "values": {
            "items": [
              {
                "value": "*"
              }
            ]
          }
        },
        {
          "key": "aws:Resource",
          "values": {
            "items": [
              {
                "value": "image/*"
              }
            ]
          }
        },
        {
          "key": "aws:Type",
          "values": {
            "items": [
              {
                "value": "image"
              }
            ]
          }
        },
        {
          "key": "aws:Account",
          "values": {
[...]

PR #187, opened 2022-08-26
[CI:TOOLING] Fix silent imgobsolete error + check status JSON

cevich:
Fixes: #186 ([Example log](https://cirrus-ci.com/task/5398202080821248))

For whatever reason, the `aws` CLI does not exit non-zero if there was
an error obsoleting an image.  Worse, it forces callers to verify a JSON
blob as well.

Fix this, and improve the script by accumulating any errors and exiting
non-zero after the entire loop finishes.  In the case of `aws`,
additionally verify the output JSON provides positive confirmation of a
successful update.

### DIFF
--- a/imgobsolete/entrypoint.sh
+++ b/imgobsolete/entrypoint.sh
@@ -6,7 +6,7 @@
 # these images are randomly selected and made obsolete.   They are also
 # marked with deletion-metadata some time in the future.
 
-set -e
+set -eo pipefail
 
 # shellcheck source=imgts/lib_entrypoint.sh
 source /usr/local/bin/lib_entrypoint.sh
@@ -185,6 +185,8 @@ then
     die 0 "Safety-net Insufficient images ($COUNT) to process ($OBSOLETE_LIMIT required)"
 fi
 
+# Don't let one bad apple ruin the whole bunch
+ERRORS=0
 sort --random-sort $TOOBSOLETE | tail -$OBSOLETE_LIMIT | \
     while read -r cloud image_name reason; do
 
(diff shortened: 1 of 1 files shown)


cevich:
@edsantiago PTAL when you get a chance, should be an easy one.

Meanwhile, I've got to actually try and figure out why _some_ EC2 images are throwing a permission error when this script runs...

edsantiago:
How did you test this? Did you? I've never used `let`, and it turns out there's an excellent reason for that. See `bash(1)`, or try this, to find out why:
```bash
$ ERRORS=0
$ let "ERRORS++"
```
(This assumes you have a color-changing prompt that will immediately let you know what happens).

cevich:
> How did you test this? Did you?

Actually I did and believed it worked, but apparently I wasn't paying enough attention or something...

cevich:
...huh, yeah it seems to work fine for me:

```bash
~/dev/automation_images (import_images|✔) $ bash
[cevich@fedora automation_images]$ F=0
[cevich@fedora automation_images]$ let "F++"
[cevich@fedora automation_images]$ echo $F
1
[cevich@fedora automation_images]$ let "F++"
[cevich@fedora automation_images]$ echo $?
0
[cevich@fedora automation_images]$ echo $F
2
```

Is this the behavior you're concerned about: `If the last arg evaluates to 0, let returns  1;  0 is returned otherwise.`?  ERRORS is incrementing, and unless it wraps at some big number should never be 0...oh wait, I see.  The first one does return 1 doesn't it.  Ugh.  Okay I'll fix it.

cevich:
Okay it's fixed.  What odd behavior, I've used `let` for ages and never noticed that return's 1 thing.  I wonder what the intent behind that design was, seems to make the call less generally-useful :disappointed:

edsantiago:
LGTM

cevich:
Thanks @edsantiago I appreciate it.

PR #188, opened 2022-08-30
[CI:TOOLING] Fix imgobsolete EC2 AMI reference by name

cevich:
All `aws` cli commands dealing with images require specifying them by
AMI ID rather than name.  When adding EC2 images into the list for
potential deprecation, only one condition used the ID, the rest used the
image name.  This will fail randomly, depending on which items are
chosen.  Fix this by always using the AMI ID when adding to the
candidate list.  Also, stick the image name-tag into the tag list for
the benefit of humans looking at logs.


cevich:
Fix worked: https://cirrus-ci.com/task/5695085420478464

Issue #150, opened 2022-07-15
`rustup .. stable` in rust images


lsm5:
Move the `rustup install/default stable` commands from netavark https://github.com/containers/netavark/pull/338 to the image builds here.

cevich:
More specifically, [move this code into](https://github.com/containers/netavark/pull/338/files#diff-db575befc56ec85f83b824430b9b62e20eb19d08b351dc743db26848639896a6R81-R89) `cache_images/fedora-netavark_packaging.sh`

PR #200, opened 2022-09-21
add dnsmasq to netavark image

baude:
The netavark-dhcp-proxy test suite needs the dnsmasq package to serve as a DHCP server.

Signed-off-by: Brent Baude <bbaude@redhat.com>

### DIFF
--- a/cache_images/fedora-netavark_packaging.sh
+++ b/cache_images/fedora-netavark_packaging.sh
@@ -27,6 +27,7 @@ INSTALL_PACKAGES=(\
     bzip2
     curl
     dbus-daemon
+    dnsmasq
     findutils
     firewalld
     gcc
@@ -51,6 +52,7 @@ INSTALL_PACKAGES=(\
     openssl-devel
     podman
     policycoreutils
+    protobuf-devel
     rsync
     sed
     slirp4netns


cache_images/fedora-netavark_packaging.sh
```diff
@@ -27,6 +27,7 @@ INSTALL_PACKAGES=(\
     bzip2
     curl
     dbus-daemon
+	dnsmasq
```
cevich, line 30: Please fix whitespace.

cevich: (No need to wait for images to build for that change)

cevich:
Thanks.  Oh, I was about to merge this but I see you added another thing.  I'll wait for the build to finish.

cache_images/fedora-netavark_packaging.sh
```diff
@@ -52,6 +53,7 @@ INSTALL_PACKAGES=(\
     podman
     policycoreutils
     rsync
+    rust-protoc-rust+default-devel
```
cevich, line 56: We're not using the packaged rust or dependencies in Fedora because they tend to be very old and outdated.  See the `"Initializing upstream rust environment."`section below.  That's where you'd want to pull something like this in (assuming it's something to do with the toolchain).

cevich:
@baude while you're fixing the `rust-protoc-rust+default-devel` issue, maybe you can also address https://github.com/containers/automation_images/issues/150 (maybe responsible for clippy findings "churn")  -  Only suggesting since it seems likely to be a highly related fix.

cache_images/fedora-netavark_packaging.sh
```diff
@@ -51,6 +52,7 @@ INSTALL_PACKAGES=(\
     openssl-devel
     podman
     policycoreutils
+	protobuf-devel
```
cevich, line 55: Is editor setup to follow the indentation style/method used in the rest of the file?

baude: evidently not

cevich:
Please feel free to merge this if the images work for you, so other PRs (here) can pick up the changes.

PR #213, opened 2022-09-29
Add STAGE build-arg support to build_push

cevich:
This is needed to support https://github.com/containers/buildah/pull/4300


cevich:
Force-push: Splitting off addition of new tests into a separate PR.

PR #225, opened 2022-10-13
Temp. remove podman package from Ubuntu Images

cevich:
A problem with broken dependency on netavark causes image build failure.
Since Podman CI doesn't need a podman package, simply remove it
temporarily.  Then sprinkle some hope and prayer around, that nobody gets
the "smart" idea to try and use this image anywhere outside of podman
CI.


cevich:
Closing...unnecessary with https://github.com/containers/podman/pull/16161 fix.

PR #228, opened 2022-10-14
Introduce Windows VM AMI with WSL enabled

n1hility:
This PR introduces the build of a Windows 2022 Server AMI image with WSL enabled for CI testing.

While this image is generically usable, it's primary purpose is to enable Podman CI to verify Windows functionality, which relies on an environment supporting WSL. Since WSL requires the use of virtualization, the image must be run on bare metal nodes. However, the image itself is built on lower-cost t3 instances since virt is only necessary to *run* WSL (installing does not require it).

While the intended target for Podman on Windows is Windows 10 and Windows 11 Desktop environments, AWS only supports the usage and licensing of Server builds. Since Windows Server 2022 and Windows 10 are derived from the same common baseline, 2022 is sufficient for testing (earlier versions (e.g., 2019) are not.

To isolate the windows building functionality, this PR introduces a new win_images Makefile target, called by a new win_images task, and a windows specific packer template and support scripts. Currently, this only builds a single image, but it could be modified in the future to use a matrix similar to base_images (it currently reuses ci/make.sh). As with other image builds, windows images can be disabled using a "no_win_server_wsl" tag.

Since Podman currently only supports x86_64 (win support of ARM is very limited at this stage), there is only a single architecture required.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -34,6 +34,7 @@ validate_task:
         - "make image_builder/gce.json"
         - "make base_images/cloud.json"
         - "make cache_images/cloud.json"
+        - "make win_images/win-server-wsl.json"
 
 
 image_builder_task:
@@ -230,6 +231,32 @@ cache_images_task:
     script: "ci/make.sh cache_images"
     manifest_artifacts:
         path: cache_images/manifest.json
+        type: application/json
+
+
+# Builds Windows server VMs for CI
+# Currently this produces a single Win 2022 Server Base image with WSL enabled
+win_images_task:
+    name: "Build VM Win-images"
+    alias: "win_images"
+    only_if: *is_pr
+    skip: *ci_docs_tooling
+    depends_on:
+        - 'validate'
+    # Packer needs time to clean up partially created VM images
+    auto_cancellation: $CI != "true"
+    stateful: true
+    timeout_in: 45m
+    # Packer WinRM communicator is not reliable on container tasks
+    gce_instance:
+        <<: *ibi_vm
+    env:
+      PACKER_BUILDS: "win-server-wsl"
+      GAC_JSON: *gac_json
+      AWS_INI: *aws_ini
+    script: "ci/make.sh win_images"
+    manifest_artifacts:
+        path: win_images/manifest.json
         type: application/json
 
 
@@ -461,6 +488,7 @@ success_task:
         - tooling_images
         - base_images
         - cache_images
+        - win_images
         - test_imgts
         - imgts
         - test_imgobsolete
--- a/win_images/win-server-wsl.yml
+++ b/win_images/win-server-wsl.yml
@@ -1,0 +1,76 @@
+variables:
+  # Naming suffix for images to prevent clashes
+  IMG_SFX:
+
+  # Allows providing handy cross-reference to the build log
+  CIRRUS_TASK_ID: "{{env `CIRRUS_TASK_ID`}}"
+
+
+builders:
+  - type: amazon-ebs
+    name: win-server-wsl
+    source_ami_filter:
+      filters:
+        name: &win_release "Windows_Server-2022-English-Full-Base*"
+        root-device-type: ebs
+        virtualization-type: hvm
+      most_recent: true
+      owners:
+        - amazon
+    # While this image should run on metal, we can build it on smaller/cheaper systems 
+    instance_type: t3.large
+    force_deregister: true  # Remove AMI with same name if exists
+    force_delete_snapshot: true  # Also remove snapshots of force-removed AMI
+    # Note that we do not set shutdown_behavior to terminate, as a clean shutdown is required
+    # for windows provisioning to complete successfully.
+    communicator: winrm
+    winrm_username: Administrator # AWS provisions Administrator, unlike GCE 
+    winrm_insecure: true
+    winrm_use_ssl: true
+    winrm_timeout: 25m
+    # Script that runs on server start, needed to prep and enable winrm
+    user_data_file: '{{template_dir}}/bootstrap.ps1' 
+    # Required for network access, must be the 'default' group used by Cirrus-CI
+    security_group_id: "sg-042c75677872ef81c"
+    ami_name: &ami_name '{{build_name}}-c{{user `IMG_SFX`}}'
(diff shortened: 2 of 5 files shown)


win_images/win-server-wsl.yml
```diff
@@ -0,0 +1,87 @@
...
+    communicator: winrm
+    winrm_username: Administrator # AWS provisions Administrator, unlike GCE 
+    winrm_insecure: true
+    winrm_use_ssl: true
+    winrm_timeout: 25m
+    # Script that runs on server start, needed to prep and enable winrm
+    user_data_file: '{{template_dir}}/bootstrap.ps1' 
+    # Required for network access, must be the 'default' group used by Cirrus-CI
+    security_group_id: "sg-042c75677872ef81c"
+    ami_name: &ami_name '{{build_name}}-w{{user `IMG_SFX`}}'
```
cevich, line 35: When bringing updated images into CI (like podman) it's convenient to be able to just update the one `IMAGE_SUFFIX` variable and have it haul everything in together.  That's not going to be possible if the name suffix uses a `w` - at least not without lots of windows-specific changes to the CI setup.  I think maybe that can be avoided with either:

* Fudge it from a `w` to a 'c' here.
* Go whole-hog and merge `win_images/*` under `cache_images` (and into `cloud.yml`).

Either way, I see no reason to treat windows as a second-class citizen here.  It's fine to include it into the rest of the bunch.  I think that also means all the existing documentation will apply to the new windows images too, which would be nice.

WDYT?

win_images/win-server-wsl.yml
```diff
@@ -0,0 +1,87 @@
...
+      - Set-Service winrm -StartupType Disabled
+      # Rerun sysprep to disable RDP and allow the AMI to be usable standablone
+      - '& ''C:/Program Files/Amazon/EC2Launch/ec2launch'' sysprep --block'
+post-processors:
+  - type: 'manifest'
+    output: '{{template_dir}}/manifest.json'  # Collected by Cirrus-CI
+    strip_path: true
+    custom_data:
+      IMG_SFX: '{{ user `IMG_SFX` }}'
+      STAGE: win
```
cevich, line 86: See my comment above.  I think we best just call this a 'cache' image proper.

cevich (commented):
This is a really good start and maybe really close to being done.  I'm glad you were able to figure out my over-complicated mess of a Makefile :grin:  The true-test will be getting this working in podman CI.  I would say we can just leave this PR open until that part is done/ready and merge them both at the same time (using a full set of fresh images built here as a final step).

.cirrus.yml
```diff
@@ -233,6 +233,34 @@ cache_images_task:
...
+    auto_cancellation: $CI != "true"
+    stateful: true
+    timeout_in: 45m
+    # Packer WinRM communicator is not reliable on container tasks
+    gce_instance:
+        <<: *ibi_vm
+        # Nested-virt is required, need Intel Haswell or better CPU
+        enable_nested_virtualization: true
+        type: "n2-standard-2"
+        scopes: ["cloud-platform"]
```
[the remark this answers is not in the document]
n1hility: yes this should be fine, nested virt isnt really needed it was just a cut and base from base_image

win_images/win-server-wsl.yml
```diff
@@ -0,0 +1,87 @@
...
+      - Set-MpPreference -DisableRealtimeMonitoring 1
+  - type: windows-shell
+    inline:
+      # Install chocolatey provisioning tool (required by cirrus)
+      - '@"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -NoLogo -InputFormat None -ExecutionPolicy Bypass -Command "iex ((New-Object System.Net.WebClient).DownloadString(''https://chocolatey.org/install.ps1''))" && SET "PATH=%PATH%;%ALLUSERSPROFILE%\chocolatey\bin"'
+  - type: powershell
+    inline:
+      # Install Git, BZ2 archive support, Go, and the MingW (GCC for Win) compiler for CGO support
+      - choco install -y git mingw archiver
+      - choco install golang --version 1.19.2 -y
```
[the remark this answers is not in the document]
n1hility: Sure I can do that

win_images/win-server-wsl.yml
n1hility: Happy either way on this. The yaml could be merged but we would need two CI tasks, unless we switch cache images to a gce instance as well. I ran into an issue with winrm communication from the container runner instance approach that cache images is currently using. I suspected it was firewall policy but did not debug it further.

cevich: The container-based tasks have some odd quirks, I'm not surprised you found one.  Two tasks is fine by me, as is switching to VMs task for cache_images.  I dunno, I don't have super strong feelings beyond the `w` -> `c` change.  It's probably not going to do much for the diff size, but might be a bit easier to maintain if the templates and scripts are all in one place?  Hmmm.  We can let this marinate.

cevich: Thought about it more.  Let's just leave the easy-thing in place (`w` -> `c`) for now.  If it makes sense later to merge all this content under `cache_images/` that can be done later.  My motivation is saving our effort for the "real" task here: getting this working in podman CI.

win_images/win-server-wsl.yml
```diff
@@ -0,0 +1,73 @@
...
+provisioners:
+  - type: powershell
+    script: '{{template_dir}}/win_packaging.ps1'
+  - type: windows-restart
+  - type: powershell
+    inline:
+      # Disable WinRM as a security precuation (cirrus launches an agent from user-data, so we don't need it)
+      - Set-Service winrm -StartupType Disabled
+      # Rerun sysprep to disable RDP and allow the AMI to be usable standablone
+      - '& ''C:/Program Files/Amazon/EC2Launch/ec2launch'' sysprep --block'
```
cevich, line 63: Out of curiosity...if we get rid of the sysprep, could that let us skirt by the 15-minute re-snowflake bull$#17 or will the CI VMs refuse to boot?

n1hility:
@cevich PTAL, podman#16257 is ready to go after this merges

PR #233, opened 2022-11-07
Support building/pushing pipglr image

cevich:
This is the Podman-in-podman GitLab Runner image added by https://github.com/containers/podman/pull/16378


cevich:
I'm going to abandon this for now in favor of https://gitlab.com/qontainers/pipglr because I cannot get our multiarch builds running. It seems there aren't binaries available and/or they're named with oddball non-standard names. Can re-open this and https://github.com/containers/podman/pull/16378 if needs and/or situation changes.

Issue #191, opened 2022-09-07
RFE: Make GCE and AWS EC2 images public


cevich:
There's no secret-sauce in our VM images.  However keeping them private excludes the community from utilizing the images for auxiliary purposes.  For example, leveraging GCE `compute_engine_instance` in their own Cirrus-CI setup.

Making images public in GCE [requires a `gcloud` command.  ](https://cloud.google.com/compute/docs/images/managing-access-custom-images#share-images-publicly), it cannot be done in packer.  OTOH, making our AWS EC2 images public is just a change to the packer templates.

Thanks to @n1hility for the suggestion and for pointing this out.

cevich:
I can't easily/safely fix this using our automation workflows for GCE because it involves granting far to permissive IAM access to the service account.  Since every PR potentially publishes new images, it's simply too large of an attack vector for me to open.

AWS Images OTOH are much easier to make public, since all I need to do is remove some metadata we set with Packer.  The AWS images are substantially identical to the GCP ones, at least for Fedora.

cevich:
Closing as "can't fix".  Opening up the GCP images requires relaxing build-time security posture more than I'm willing.  Opening up AWS image will swiftly run into their "max 5 public AMIs" limit.

PR #238, opened 2022-12-07
Make AWS AMI's accessible by the public

cevich:
Fixes: #191

On at least two occasions I can remember, users outside the containers-team have requested access to our CI VM images. Unfortunately doing this for the GCE images has some security-interest conflicts with our build automation.  However, making the AWS AMI's public is rather easy to do.  In fact it's the default unless they are specifically locked down.  Simply remove the private restrictions and allow access to `"all"` AWS groups.


cevich:
Apparently there's a limit of --> 5 <-- public AWS EC2 images 🫣

PR #241, opened 2022-12-14
netavark: install mandown into image

Luap99:
mandown is required to convert a markdown document into a man page.

see https://github.com/containers/netavark/pull/534


Luap99:
@cevich PTAL

cevich (approved):
LGTM - I restarted the two failed jobs, they were simply Cirrus-CI farts.

cevich:
Re-ran `fedora-aws Base Image` flake.

Luap99:
Ok now it worked I had to set CARGO_HOME for the install command.

cevich:
Ahh this makes sense since the environment doesn't pass through the `$SUDO`.  Good catch and the fix LGTM.

Issue #243, opened 2023-01-09
Base-image build ssh-disconnect


cevich:
While building a new prior-fedora (F36) base-image, it appears as if ssh is [disconnecting during a package update](https://cirrus-ci.com/task/5424275562168320):

```
...cut...
==> prior-fedora: Using SSH communicator to connect: 127.0.0.1
==> prior-fedora: Waiting for SSH to become available...
==> prior-fedora: Connected to SSH!
==> prior-fedora: Pausing 10s before connecting...
==> prior-fedora: Waiting for SSH to become available...
==> prior-fedora: Connected to SSH!
==> prior-fedora: Provisioning with shell script: /tmp/packer-shell3026262523
==> prior-fedora: Uploading /tmp/cirrus-ci-build/ => /tmp/automation_images/
==> prior-fedora: Provisioning with shell script: /tmp/packer-shell1652337550
    prior-fedora: Disabling periodic services that could destabilize automation:
    prior-fedora: Banishing cron (ignoring errors)
    prior-fedora: Banishing crond (ignoring errors)
    prior-fedora: Banishing atd (ignoring errors)
    prior-fedora: Banishing apt-daily-upgrade (ignoring errors)
    prior-fedora: Banishing apt-daily (ignoring errors)
    prior-fedora: Banishing fstrim (ignoring errors)
    prior-fedora: Banishing motd-news (ignoring errors)
    prior-fedora: Banishing systemd-tmpfiles-clean (ignoring errors)
    prior-fedora: Banishing update-notifier-download (ignoring errors)
    prior-fedora: Banishing mlocate-updatedb (ignoring errors)
    prior-fedora: Warning: Automation library not found. Assuming it's not yet installed
    prior-fedora: Fedora 36 - x86_64                               18 MB/s |  81 MB     00:04
    prior-fedora: Fedora 36 openh264 (From Cisco) - x86_64        2.0 kB/s | 2.5 kB     00:01
    prior-fedora: Fedora Modular 36 - x86_64                      2.9 MB/s | 2.4 MB     00:00
    prior-fedora: Fedora 36 - x86_64 - Updates                     16 MB/s |  31 MB     00:01
==> prior-fedora: Provisioning step had errors: Running the cleanup provisioner, if present...
==> prior-fedora: Deleting output directory...
[...]

PR #244, opened 2023-01-09
Fix ssh disconnect during package update

cevich:
For some reason, a package update is causing ssh to disconnect.  Looking at the container-image build, the following pacakges ran scripts:

* elfutils-default-yama-scope-0.188-3.fc36.noarch
* tpm2-tss-3.2.1-1.fc36.x86_64
* libtasn1-4.18.0-2.fc36.x86_64

Since debugging at this stage of building images is difficult, simply disable the package update.  Another update will run during the cache-image stage, and hopefully not fail in a similar way.

Ref: https://github.com/containers/automation_images/issues/243


flouthoc (approved):
LGTM

cevich:
I'm going to need to manually run the build to debug this further.  That should let me better see what's happening, maybe it's hitting an OOM on the VM or something.

cevich:
Okay i think I reproduced the problem locally, QEMU is getting OOM-killed during the package update.  I think I see this happening in Cirrus as well, memory usage peaks for a minute or three before the error occurs.  My local devel. env. is close but not 100% the same as Cirrus, though the differences are probably not important.

I have some ability to set QEMU args, but since this has never been a problem in past years, I'm wondering if maybe `dnf` is to blame, and where a fix could be better targeted.

cevich:
Update: Looked closer, it's not QEMU getting killed it's dnf:

```
prior-fedora: Fedora 36 - x86_64 - Updates                     16 MB/s |  31 MB     00:01
[   71.007137] gmain invoked oom-killer: gfp_mask=0x1100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
[   71.009203] CPU: 0 PID: 627 Comm: gmain Not tainted 5.17.5-300.fc36.x86_64 #1
[   71.010653] Hardware name: Red Hat KVM, BIOS 1.16.0-3.module_el8.7.0+1218+f626c2ff 04/01/2014
[   71.012292] Call Trace:
...cut...
[   71.203304] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/user.slice/user-0.slice/session-3.scope,task=dnf,pid=1626,uid=0
[   71.207640] Out of memory: Killed process 1626 (dnf) total-vm:1181752kB, anon-rss:53440kB, file-rss:4kB, shmem-rss:0kB, UID:0 pgtables:2076kB oom_score_adj:0
```

cevich:
Updated description, force-pushed a more informed workaround.

PR #247, opened 2023-01-17
Human-readable date-based image suffixes

cevich:
Historically, the ID used to identify a set of images was taken from the
Cirrus-CI build ID.  This was done to make auditing easier, since one
can easily retrieve the build logs using the ID.  However, there are
many disadvantages to using the build id:

* It's not human-readable, making it difficult to ascertain exactly when
  the images were built.
* It's not guaranteed to be incremental, and therefore cannot be
  utilized as a "version".
* It doesn't convey helpful information like when it was produced, or
  which release of Fedora is included in the set.

For these and other reasons, switch to a simple date-based image suffix
encoded in a repository file.  Include a UTC-baesd timestamp to avoid
value collisions, and a string identifying what OS's releases were built.


cevich:
Right...`.cirrus.yml` has references to `$IMG_SFX` all over the place, and there's no way to pull the value from the Makefile :cry:

cevich:
@edsantiago This is not urgent, but I could use a second pair of eyes early on.

Big-picture: My eventual goal is to have renovate manage opening vm-image update PRs for us.  For that to work, we need image-suffixes resembling version-numbers that sort properly, and can be `git tag` into this repo.   Before you ask - "no" the current values are not guaranteed to be incremental.

So this PR is a first-step in that direction...but as usual, it's complicated :disappointed: 

The major change here so far, is that Cirrus will read a $IMG_SFX "version" value in from a file in the repo.  The file can be generated (and updated) by `make IMG_SFX`. However, since it's date-based (for human readability), there can be clashes[0].  I added a `v0` field that can be incremented to help avoid them, but this is less than ideal[1].

In short, with this PR, the new "build a new set of images" workflow would go like:

1. Run `make IMG_SFX`.
2. Manually verify there are/were no other open image-update PRs that used the same suffix value - this could be an automated check eventually, but that's for later.
3. If there is a clash, edit `IMG_SFX` file and bump the `v0` number to prevent it.
4. Push PR and wait.
5. Assuming success, github-action prints table of built images.
6. Update downstream .cirrus.yml to use the new suffix,
   e.g. `IMAGE_SUFFIX = c230118v0-f37p36u2204`

Does this sound mostly sane so far?

Do you think it's helpful to include the OS revisions in the ID?

---

`[0]`Note: A `IMG_SFX` clash is potentially really very bad.  At best it will clobber existing images from somebody else's PR.  At worst it will break somebody's active (still building) images or images already in use by downstream CI (eek!).

`[1]`Note: There are character and size limitations to the IMG_SFX value.  I'd prefer to keep it human readable if possible, but an option could be to swap the OS-versions part for a  short git-commit hash to prevent clashes.

Makefile
```diff
@@ -115,6 +114,11 @@ help: ## Default target, parses special in-line comments as documentation.
 	@grep -E '^[[:print:]]+:.*?## .*$$' $(MAKEFILE_LIST) | \
 		awk 'BEGIN {FS = ":(.*)?## "}; {printf $(_HLPFMT), $$1, $$2}'
 
+# N/B: There are length/character limitations in GCE for image names
+.PHONY: IMG_SFX
+IMG_SFX:  ## Generate a new date-based image suffix into the file IMG_SFX
+	$(file >$@,$(shell date +%y%m%d)v0-f$(FEDORA_RELEASE)p$(PRIOR_FEDORA_RELEASE)u$(subst .,,$(UBUNTU_RELEASE)))
```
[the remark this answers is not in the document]
cevich: SGTM.  Looks like the limit is 63-characters.  Since the timestamp is being kept in a user-generated file, let me try adding hours/minutes/seconds as well.  That will avoid the whole "woops I overwrote your image" problem.

cevich:
force-push change summary:

* Clarified README.md changes further
* Added '-' separator between distro versions in IMG_SFX format
* Added hours/minutes/seconds value to IMG_SFX format
* Moved validation commands into `validate.sh` w/ helpful failure messages.

cevich:
w00t! :tada:

Luap99:
I find it very hard to read this custom date format. Any reason to not just use the iso 8601 format? Also looks like the time is in the local timezone? I would prefer UTC.

Lastly would it make sense to encode the automation_images PR number in that ID? In case something broke it will be easy to find the correct PR here.

cevich:
@Luap99 Using UTC is a good suggestion. Thanks.

I'm fairly limited in the character set and length for the image names.  It's about 45 lower-case letters, numbers, and '-'.  That's it.  So I'm afraid a delimited date/time wouldn't look much better:

`23-01-19-13-13-09-f37-f36-u2204`

what about if I stuck a full year, and `t` in there in they style of 8601, so:

`2023-01-19t13-13-09-f37-f36-u2204`

any better?

cevich:
Also keep in mind, it's not expected for humans to be reading/decoding this ID very much.  Maybe once a month or so on average, if that.  So I think some amount of ugliness is probably okay.

cevich:
Hmmm.  I think all the `-` actually make it harder to read.  What about:

`20230120-151738-f37f36u2204`  (eventually will be `d11` on the end)

Luap99:
> @Luap99 Using UTC is a good suggestion. Thanks.
> 
> I'm fairly limited in the character set and length for the image names. It's about 45 lower-case letters, numbers, and '-'. That's it. So I'm afraid a delimited date/time wouldn't look much better:
> 
> `23-01-19-13-13-09-f37-f36-u2204`
> 
> what about if I stuck a full year, and `t` in there in they style of 8601, so:
> 
> `2023-01-19t13-13-09-f37-f36-u2204`
> 
> any better?

`20230120T141842Z` would be also work for me and iso 8061

I don't really need the dashes, the full year alone is enough for my brain to recognize this as date.

cevich:
SGTM

cevich:
Upper-case characters aren't supported, but I still think this works okay:

`20230120t152650z-f37f36u2204`

The real trick will be seeing if I can teach renovate how to interpret that as a version number.

cevich:
> I really like the idea behind this.

Ya, me too.  I think it was actually @edsantiago that suggested it a few years ago :relaxed: The thing that's motivating me is: Renovate will be able to post image-update PRs, just based on tagging one of these "versions" in this repo.  I'm working on a POC in containers/automation_sandbox right now, as I write this.

cevich:
W00t, it's working: https://github.com/containers/automation_sandbox/pull/97

ci/validate.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+# Verify new IMG_SFX value always sorts later than previous value.  This prevents
+# screwups due to local timezone, bad, or unset clocks, etc.
+cd $REPO_DIRPATH
+# Automation requires the date and time components sort properly
+# as if they were version-numbers regardless of distro-version component.
+new_img_ver=$(awk -F '-' '{print $1"."$2}' ./IMG_SFX)
+# TODO: Conditional checkout not needed after the PR which added IMG_SFX file.
+if git checkout ${CIRRUS_BASE_SHA} IMG_SFX; then
```
edsantiago, line 43: This `checkout` makes me uncomfortable: it leaves behind a bad file which may bite someone down the line. It's probably okay, ... unless someone adds a new YAML step after `ci/validate.sh`.

Would you consider using
```bash
v_prev=$(git show ${CIRRUS_BASE_SHA}:IMG_SFX 2>&1 || true) 
if ! [[ $v_prev =~ "exists on disk, but not in" ]]; then
    ...
```

edsantiago (approved):
High-level: wow, fantastic work, thank you. This is beautiful and will be much appreciated.

Low level: whewwwwww, thanks for your patience. A lot to learn and understand. I have one (possibly invalid) concern, and one question.

.cirrus.star
```diff
@@ -0,0 +1,11 @@
+#
+# Lang. ref: https://github.com/bazelbuild/starlark/blob/master/spec.md#contents
+# Impl. ref: https://cirrus-ci.org/guide/programming-tasks/
```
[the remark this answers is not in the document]
cevich: Oh man you're not kidding.  I spent a few days down the hole, then deleted everything and just made it read from the IMG_SFX file :grin:

ci/validate.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Variable is defined by Cirrus-CI at runtime
+# shellcheck disable=SC2154
+if ! git diff --name-only ${CIRRUS_BASE_SHA}..HEAD | grep -q IMG_SFX; then
+  die "Every PR must include an updated IMG_SFX file.
+Simply run 'make IMG_SFX', commit the result, and re-push."
+fi
+
+# Verify new IMG_SFX value always sorts later than previous value.  This prevents
+# screwups due to local timezone, bad, or unset clocks, etc.
```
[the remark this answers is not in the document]
cevich: Oh ya, there's a setting for that: "Require branches to be up to date before merging".  I'm hesitant to turn it on now, because it will force people to constantly rebuild their images, even when doing so would be anti-helpful:  PRs in this repo. tend to take a long time to merge, and constantly rebasing them will mean constantly updating the IMG_SFX in downstream CI PR's :confounded: 

So for now, I think this might be a temporary problem until the general matter of image-update workflows is refined (and documented better).  Evenso, ultimately what will matter the most is the order IMG_SFX values are tagged in the repo.  That's the action that will trigger renovate to open image-update PRs for downstream CI.  Anyway...that's a future problem.

ci/validate.sh
cevich: I could but...this is hurting my brain at the moment :grin:   I think my intention was (after this merges) to immediately remove the `if` so it always does the checkout.

Are you suggesting that at the end, after validating (good or bad), the script should `git checkout HEAD IMG_SFX` to make the repo. pristine?  Assuming so, I could live with that.  Opened https://github.com/containers/automation_images/pull/251

cevich:
> whewwwwww, thanks for your patience

No worries, as always I appreciate the oversight *MUCH* more than I dislike waiting :smile:

ci/validate.sh
edsantiago: No, my suggestion was never to do the unsafe checkout in the first place...

edsantiago: But yeah, that can be done in a followup PR. If you want to merge this, I'm OK with that.

cevich: Okay, made a note of this in followup PR.

PR #248, opened 2023-01-17
[WIP] Use github's check-suite ID for images

cevich:
In order to treat image-suffixes like version-numbers, they must be both unique, and monotonically increasing from one PR to another. Unfortunately there are no such guarantees for the Cirrus Build ID values - used prior to this commit.  A number of tests and checks require the suffix value to be defined from the `.cirrus.yml` configuration.  Switch to using the `GITHUB_CHECK_SUITE_ID` which is guaranteed to be unique and ever-increasing.


cevich:
Won't work.

PR #252, opened 2023-02-02
[CI:TOOLING] Validate: Catch attempted re-use of IMG_SFX

cevich:
IMG_SFX values need to change for every image build, even within the
same PR.  Attempt to catch re-use of a suffix before starting the lengthy
build process (which will eventually fail on a duplicate).  Check the
`imgts` image simply because it builds very early in cirrus-ci and cannot
be skipped with a "no_*" PR label.

Also - enforce IMG_SFX change rules when building in `[CI:TOOLING]`
mode, since those container images push with a IMG_SFX tag.  It would be
unfortunate if an unrelated PR clobbered in-use tooling images w/ a
breaking-change.


cevich:
[Example validation failure](https://cirrus-ci.com/task/5010655246286848) due to no IMG_SFX file change.

cevich:
[Example validation failure](https://cirrus-ci.com/task/5692745679896576) due to IMG_SFX re-use.

cevich:
[Example validate pass](https://cirrus-ci.com/task/6696755991937024) w/ TOOLING-only update (magic string) + new IMG_SFX value.

PR #254, opened 2023-02-09
[DO NOT MERGE] Enable Fedora Rawhide

lsm5:
Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>


base_images/fedora_base-setup.sh
```diff
@@ -16,6 +16,11 @@ REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
 # shellcheck source=./lib.sh
 source "$REPO_DIRPATH/lib.sh"
 
+# Switch to Rawhide
```
lsm5, line 19: @cevich I see there's a `shellcheck disable=SC2154` here. Do I need to move it before this part ?

base_images/fedora_base-setup.sh
```diff
@@ -16,6 +16,12 @@ REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
 # shellcheck source=./lib.sh
 source "$REPO_DIRPATH/lib.sh"
 
+# Update to Rawhide
+# NOTE: Should not break Fedora releases
+if [[ -z BUILD_RAWHIDE ]]; then
```
cevich, line 21: I think maybe you want a `-n` test here.

base_images/fedora_base-setup.sh
cevich: You need to disable SC2154 for any env. var. that's defined elsewhere, like `$CIRRUS_foo` for example.

cevich:
Taking this over as https://github.com/containers/automation_images/pull/273

PR #270, opened 2023-04-10
[CI:TOOLING] Fix removal of active AWS images

cevich:
Periodically image `LastLaunched` timestamps are examined.  When found to be more than 30-days in the past, images have a deprecation date set on them 30-days in the future.  However during the periodic scans, if the `LastLaunched` time is updated, the deprecation date isn't cleared.

Worse, a previous bug had been incorrectly updating deprecation dates on active images.  That bug was fixed, but the deprecation status was never cleaned up.  This has been corrected manually by removing the deprecation statuses of all EC2 images.

Fix both the periodic (daily) obsolete scans, and CI-job level (imgts) activity scans such that the deprecation status of any active image is cleared.


cevich:
Note to me:
- [ ] Manually test imgts
- [ ] Manually test imgobsolete
- [ ] Re-enable cirrus-cron job

PR #275, opened 2023-04-21
Update to Fedora 38

cevich:
Also twiddle a few other bits and bobs needing minor touches here and there.


cevich:
Note to me: https://bugzilla.redhat.com/show_bug.cgi?id=2171472 fixed by https://bodhi.fedoraproject.org/updates/FEDORA-2023-dcf5ef9eda

PR #279, opened 2023-05-10
Use an ssh-agent in get-ci-vm.sh

nalind:
Use an ssh-agent to cache keys in the container, so that if the user's gcloud key is encrypted, we don't have to prompt them for the passphrase several times.


cevich:
Oh cool!  Sorry I didn't notice this PR, my bad.  This is a welcome addition.  Thanks @nalind!

PR #286, opened 2023-07-31
[skip-ci] [CI:TOOLING] Automatic termination of EC2 VMs

cevich:
Around the time of this commit, an annoyingly steady stream of EC2 orphans were reported to Cirrus-support.  They've taken actions to resolve, but the failure-modes are many and complex.  Since most of the EC2 instances are rather expensive to keep needlessly running, and manual cleanup is annoying, enhance the monitoring script try to attempt termination automatically.

This isn't perfect, it's possible for the script to break in strange ways and it's not practical to check for all of them.  Instead, include helpful indications in the monitoring e-mail regarding what was attempted.


cevich:
@edsantiago if you could give this a quick sanity-check for now I'd appreciate it.  I still need to rig up a manual test to confirm it's not going to try and delete every VM in the cloud.

orphanvms/_ec2
```diff
@@ -77,6 +78,33 @@ for instance_index in $(seq 1 $(jq -e 'length'<<<"$simple_inst_list")); do
...
+        dbg("Attempting to terminate instance $instance (Name=$name)")
+
+        # Operation runs asynchronously, no error reported for already terminated instance.
+        # Any stdout/stderr here would make the eventual e-mail unreadable.
+        if ! termout=$(aws ec2 terminate-instances --no-paginate --output json --instance-ids "$instance" 2>&1)
+        then
+            # Catch rare TOCTOU race, instance was running, terminated, and pruned while looping.
+            # (terminated instances stick around for a while until purged automatically)
+            if [[ "$termout" =~ InvalidInstanceID ]]; then
+                line+=" auto-termination: instance vanished"
```
edsantiago, line 96: I think a separator would be really nice here, to avoid a run-on:
```diff
- ... running X days auto-termination ETC ...
+ ... running X days; auto-termination ETC ...
```
...but it's probably not worth a re-push.

edsantiago (approved):
LGTM, one usability nit. Thanks for getting to this so quickly.

orphanvms/_ec2
cevich: Oh okay, thanks.  If I need to re-push I'll add that.

cevich:
I haven't and am not likely to get a chance to properly test this.  Also (so far) the leaking VMs appear to have stopped.  Closing this for now, will re-open if it becomes necessary.

edsantiago:
LGTM with my usual caveat, that is, I have no experience with this environment and no way to test it.

PR #294, opened 2023-08-16
Replace clobbered ec2 import image

cevich:


cevich:
@lsm5 `c20230816t191118z-f38f37d13` is ready for ya.

PR #296, opened 2023-08-22
Obscure non-cache image IDs in pr-comment table

cevich:
All built images are included in the build-table added as a PR comment
to be helpful for reference and possible debugging.  However, it's
unhelpful if a human accidentally tries to deploy a non-cache image ID
into CI somewhere.  Those images are never to be used outside of very
special-case situations.  Obscure non-cache image IDs in the table to
prevent accidents.


.github/actions/bin/create_image_table.py
```diff
@@ -40,46 +40,58 @@ def stage_sort(item):
...
+    # to have them in the list in case some debugging is needed.
+    if item["stage"] != "cache":
+        image_suffix = "do-not-use"
+    lines.append(
+        "|*{0}*|[{1}]({2})|`{3}`|\n".format(
+            item["stage"],
+            item["name"],
+            "{0}/{1}".format(url, item["task"]),
+            image_suffix,
+        )
```
cevich, line 76: @edsantiago this is the main change, everything else is automated-style updates.  Does this wording seem okay?

edsantiago: I think that makes sense, thanks. I'll wait to see the table though.

Quick plea: the order of commits (change, then reformat; instead of reformat, add new code) makes it hard to review. Next time, could you do cleanup first, then new code? Or add new code using the approved format? TIA.

cevich: Aye, thanks.  I thought and struggled with that, deciding 'good enough' would be to make it a separate commit.  Though I trust your experience that the commit order matters too.  

Python lint/style tooling is annoying in that the standards seem to be a constantly moving target. Otherwise I'd have included it in the validation checks here :confused:

cevich:
Garrrh!  This can't be tested in a PR.  Have I mentioned how much I enjoy working with GHA-workflows?

edsantiago:
Oh..... duh. The new code doesn't take effect until --- yep, what you just said.

Luap99:
Thanks, I think my brain should be smart enough to not copy-paste `do-not-use` :grin:

cevich:
NP.  Sorry for the bad table layout. The code that makes it is quite full of rot, over-complexity, and neglect.  I'm trying really hard to not look at it until it becomes truly completely broken :laughing:

PR #297, opened 2023-08-22
Test PR #296

cevich:


edsantiago:
Hey @cevich as long as there's a new image-build PR, can we do something useful with it? Pasta is broken (see https://github.com/containers/podman/pull/19699), is there a dnf way to say "give me the older version"?

cevich:
> Hey @cevich as long as there's a new image-build PR, can we do something useful with it?

We can...though terminating a build in progress definitely leaks VMs.

> is there a dnf way to say "give me the older version"?

I'd say we could just reference a specific version in the package list, but it's generic.  So the other Fedora releases would break or break soon.   I s'pose a complex/ugly conditional for all-three Fedora releases could be used.

Hmmm :thinking: 

There's a `downgrade` sub-command, and it appears to still be in dnf5 (used by rawhide).  I'm afraid that may lead to some "interesting" future debugging if the `FIXME` change is never reverted.

Hmmm :thinking: 

[Just read your comment](https://github.com/containers/podman/pull/19699#issuecomment-1688769287), that's odd, it's **only** affecting F38 and not rawhide?

cevich:
Ahh, it's F38 and rawhide.  Hmmmm, maybe a `!prior-fedora` conditional could be used then.

cevich:
Unless you understand the problem differently, I think maybe it's best to wait for @sbrivio-rh to chime in.  Maybe there's already a fix and this build will pick it up.  But if the problem is ugly/nasty/sticky then skipping the tests in Paul's PR may be a better bet.

cevich:
Woops, we cross-posted:  I'm fine either way.  There's a chance my PR-table change is broken and I need to tweak this PR anyway.

cevich:
Ref: [workflow log](https://github.com/containers/automation_images/actions/runs/5943721411)

PR #302, opened 2023-09-19
Bump to Fedora 39

edsantiago:


edsantiago:
@cevich do you need to do some magic to allow this to run?

cevich:
Oh that's unexpected.  But good, I'm glad it's working.  That's meant to keep the build secrets  safe(r).  Yeah let me add you to the list...

cevich:
...okay should work now (probably when you re-push).  FWIW (and in case it breaks) I also added another check for this in https://github.com/containers/automation_images/commit/e53f780ac4963e85e7a0bea6123fee9f460e90a4 since it's really easy to accidentally change this setting.  It's not clear on the settings page if it's on or off.

cevich:
Yay!

In case some calamity befalls me or my podman PR, this commit is needed to support (already merged) updates to the automation library: https://github.com/containers/podman/pull/19751/commits/07110cbb9dc6ed85c5b86ede453fa15d938235be  I'm hoping my PR makes it in before this one, but...you never know :worried: 

Same for Buildah.  And this buildah update is related too https://github.com/containers/buildah/pull/5007 Sorry I don't have a good way to keep track of all these dependent moving parts :disappointed:

edsantiago:
Don't worry, I'll wait for yours to merge. It's going to take me a long time anyway to get these builds, then test them in CI.

Thank you!

Makefile
```diff
@@ -17,11 +17,11 @@ if_ci_else = $(if $(findstring true,$(CI)),$(1),$(2))
...
-export PRIOR_FEDORA_RELEASE = 37
+export FEDORA_RELEASE = 39
+export PRIOR_FEDORA_RELEASE = 38
 
 # This should always be one-greater than $FEDORA_RELEASE (assuming it's actually the latest)
-export RAWHIDE_RELEASE = 39
+export RAWHIDE_RELEASE = 40
 
 # See import_images/README.md
 export FEDORA_IMPORT_IMG_SFX = $(_IMPORT_IMG_SFX)
```
cevich, line 27: oh man...It just occurred to me the AWS builds are probably going to break (or worse, silently use F37/38) unless the import-images process is run.  There are no usable Fedora-beta images in AWS, so they need to be "imported".

Ugggg!  Damn...that's an ugly process and I'm afraid of subjecting you to it.

I mean, you're welcome to try, maybe it will go smoothly?

It's (local) networking intensive, needs a magic key file (in BitWarden), and a devel container (from Makefile).  But the REALLY ugly part is just past all that.  In this really stupid, annoying, asynchronous, and flaky cloud-tool that "converts" images into AMIs (slightly more than just images).  When it breaks, it's impossible to debug and "start over" is about 45 minutes away :sweat: 

Let me see if I can squeeze that in tomorrow morning.  Sorry I totally forgot about this...I think it gave me PTSD and I blocked it out :cursing_face: Sorry man :disappointed_relieved:

edsantiago:
Don't sweat it - I'll give that process a try, in the interests of spreading the ~~pain~~ knowledge.

Makefile
cevich: Yay, got lucky and it worked on the second try.  This small change will allow this PR to continnue:

```diff
diff --git a/IMPORT_IMG_SFX b/IMPORT_IMG_SFX
index 700099f..aa1f708 100644
--- a/IMPORT_IMG_SFX
+++ b/IMPORT_IMG_SFX
@@ -1 +1 @@
-20230816t191121z-f38f37d13
+20230921t135305z-f39f38d13
```

cevich:
> Don't sweat it - I'll give that process a try, in the interests of spreading the pain knowledge.

It actually may be worth while (long-term) to give it a spin.  It downloads and uploads two fedora cloud qcow2 images, from inside a CentOS container image.  So maybe run it from a cloud VM or somewhere with lots of available bandwidth.  For reference, the commands I used are documented in some more detail in the `import_images/README.md`:

```
$ make IMPORT_IMG_SFX
$ make image_builder_debug GAC_FILEPATH=/dev/null \
    AWS_SHARED_CREDENTIALS_FILE=$HOME/.aws/credentials
[root@863844a17bdb automation_images]# make import_images  # failed
[root@863844a17bdb automation_images]# make import_images  # worked
```

You know it worked when the Makefile gives two `Image import(s) successful!` outputs (one for the x86_64 image and another for arm64.

When it breaks (as it does frequently), depending on where/how it failed, you sometimes can 
just re-run the `import_images` target and it will work (as it did for me).  Other times it's easier to `make IMPORT_IMG_SFX` and re-run the `import_images` target.  Of course, if you have ideas on how to improve this workflow I'm open.

---

For ref:  The crux of 99% of the flaky-behavior happens on the AWS-end of the asynchronous loop (in Makefile) starting with `$(AWS) ec2 import-snapshot...` and polled by `bash import_images/wait_import_task.sh`.  Either a timeout will happen, a "failed to convert", or "bad source" type errors.  Those are all lies, because (assuming no name conflicts) re-running the exact same `ec2 import_snapshot` command may work perfectly the second or third time.  Sometimes it will partially work, leaving a snapshot or broken AMI or two behind, those will cause conflicts unless they're removed or a new import ID is generated.

cevich:
Yay...well, at least until the next step :rofl:

cevich:
@edsantiago I'd like to merge https://github.com/containers/automation_images/pull/304 which has an older `IMG_SFX` value WRT this PR.  Nevertheless it will cause a conflict, but I'm happy to ignore that and merge this when ready. I'm fine keeping this open for a while too, in case you think you need to do a rebuild for some reason.

edsantiago:
@cevich this isn't even close to working; see https://cirrus-ci.com/build/4823192980684800. I will need to rebuild, probably several times. So yes, please proceed with any and all other merges you need to make.

edsantiago:
Incidentally... I can't run the `make image_builder_debug` step. It requires `.aws/something` which I do not have and can't find documentation on how to create.

cevich:
> please proceed

Okay, that works too, thanks.

> Incidentally... 

That's good feedback actually.  The contents of that come from creating an "access key" as part of the AWS console account creation process.   You have one from over a year ago, it's one of those "shows on the webpage once" things.  I can make you a new one if you like and send you the details.

cevich:
> the directory/file structure for ~/.aws

Ahh, yes that may be obscured in the AWS docs (as they seem to enjoy doing).  Fortunately it's a really simple ini-style format:

```
[default]
aws_access_key_id = <Your Value Here>
aws_secret_access_key = <Your Value Here>
region = us-east-1
```

You might need to add a `output = json` line too.  I can't remember if the import-images tooling sets that or not.  It definitely expects JSON output.

edsantiago:
Successful podman CI run here:  https://github.com/containers/podman/pull/20162

...except for Windows, which I guess will require a respin, which is beyond my abilities.

edsantiago:
Successful podman CI run here: https://github.com/containers/podman/pull/20162

I think this is ready for review and merge. It is unpleasant, but it produces working VMs. The kludge code can be removed in the next followup PR.

cache_images/fedora_packaging.sh
```diff
@@ -206,6 +206,29 @@ DOWNLOAD_PACKAGES=(\
...
+# FIXME: 2023-09-26: emergency upgrade to fix chmod-symlink bug
+if [[ $(date +%Y%m) -gt 202309 ]]; then
+    echo ""
+    echo "FATAL FATAL FATAL: REMOVE THIS TEMPORARY WORKAROUND"
+    echo
+    exit 1
+fi
+source /etc/os-release
+# shellcheck disable=SC2154
+if [[ $VERSION_ID -ge 38 ]]; then
```
cevich, line 218: This is going to override on rawhide also.  I'm assuming that's desired.

---

For future ref.: The distro name and release are defined by both `lib.sh` (here) and (elsewhere) in the common automation library.  Something like:

```
OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | tr -d '.')"
OS_RELEASE_ID="$(source /etc/os-release; echo $ID)"
OS_REL_VER="$OS_RELEASE_ID-$OS_RELEASE_VER"
```

In this script's context, there's also `$PACKER_BUILD_NAME` which can be used to distinguish Fedora vs Rawhide.

cache_images/rawhide_setup.sh
```diff
@@ -17,7 +17,7 @@ source "$REPO_DIRPATH/lib.sh"
 req_env_vars PACKER_BUILD_NAME
 
 # Going from F38 -> rawhide requires some special handling WRT DNF upgrade to DNF5
-if [[ "$OS_RELEASE_VER" -eq 38 ]]; then
+if [[ "$OS_RELEASE_VER" -ge 38 ]]; then
```
cevich, line 20: good catch.

cevich (approved):
Thanks for all the `egrep` fixes.  I thought I had attended to those, apparently not.

PR #305, opened 2023-09-26
[CI:DOCS] Improve import-image docs

cevich:
Ref: https://github.com/containers/automation_images/pull/302#issuecomment-1735721791

### DIFF
--- a/import_images/README.md
+++ b/import_images/README.md
@@ -23,8 +23,18 @@ human:
   [EC2 snapshot import process](https://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-import-snapshot.html).
 * You are in possession of an AWS EC2 account, with the [IAM policy
   `vmimport`](https://docs.aws.amazon.com/vm-import/latest/userguide/required-permissions.html#vmimport-role) attached.
-* Both "Access Key" and "Secret Access Key" values set in [a credentials
+* You have "Access Key" and "Secret Access Key" values set in [a credentials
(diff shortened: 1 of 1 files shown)


import_images/README.md
```diff
@@ -23,8 +23,9 @@ human:
   [EC2 snapshot import process](https://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-import-snapshot.html).
 * You are in possession of an AWS EC2 account, with the [IAM policy
   `vmimport`](https://docs.aws.amazon.com/vm-import/latest/userguide/required-permissions.html#vmimport-role) attached.
-* Both "Access Key" and "Secret Access Key" values set in [a credentials
+* You have "Access Key" and "Secret Access Key" values set in [a credentials
```
edsantiago, line 26: That's a huge complicated long page. Is there any chance you could just say:
> Create ~/.aws/PATH with the following format: blah blah = <YOUR ACCESS KEY> \n blah2 = <SECRET KEY>`

...so that (1) the string "slash dot aws" is greppable by someone doing a quick search, and (2) that person can then copy-paste with appropriate substitutions? Keep the link, of course, but please don't make me have to read it except in a dire emergency.

cevich: Yeah, that might be a good idea.  When you setup the key, there's a link to download the ini-file.  I believe I wrote these docs with that in mind.  But yeah, the AWS docs are anything but simple, I'll add something quick here...

PR #307, opened 2023-09-26
Add hyperv to windows image

n1hility:
Also adds an example user data file to reenable rdp


n1hility:
@cevich here you go^

n1hility:
@cevich  looks like the windows build worked

cevich:
Thanks Jason.  I need to circle around WRT a crun package in the Fedora rawhide images.  Assuming that aligns, I'll merge this and put it to use.

PR #309, opened 2023-09-27
Update windows CI VMs for hyper-v machine testing

cevich:
Ref: https://github.com/containers/podman/pull/20177

In addition to updating mingw and golang, this moves the
installation of .Net and wixtoolset here instead of at CI runtime.
The windows packer-configuration was updated to operate more
consistently with how things are done in Linux WRT calling scripts.
Along with some file renames and other cosmetic changes, the windows
build timeout was increased since the extra packages seem to
place it right on the edge of the former value.


cevich:
Nope, still `1.9-1` in rawhide.

edsantiago:
I already have the desired crun in #302, it's just hand-downloaded via kludge.

cevich:
> Note that crun is now

K.  I'm hoping your PRs can be merged soon, and then I can re-purpose this one for windows podman-machine related updates.

win_images/win_packaging.ps1
```diff
@@ -25,10 +49,16 @@ Set-ExecutionPolicy Bypass -Scope Process -Force
...
+# Install basic required tooling.
+#   psexec needed to workaround session 0 WSL bug
+#   mingw version 11.2, later versions may be incompatible with CGO on some versions of golang
+retryInstall git archiver psexec golang mingw@11.2.0; CheckExit
+
+# Update service is required for dotnet 3.5 (dep of wix)
+Set-Service -Name wuauserv -StartupType "Manual"; CheckExit
+
+# Install wixtoolset (and .Net) for installer build & test.
+retryInstall wixtoolset; CheckExit
```
cevich, line 61: @n1hility when you have a sec.  I'm trying to move the mingw/wix install from podman CI to here, to try and save 5-10 minutes of podman-developer time.  Now I'm wondering if you had it setup that way for a reason?

The .Net install [seems to be breaking here fairly hard](https://cirrus-ci.com/task/5559138445426688).  If it's an easy fix, I'm eager to know, otherwise I'll just move it back into podman.  TIA.

n1hility: @cevich sorry I missed this message. I'm not sure why the package is failing. I would try two things, one putting the Set-Service higher before any installs, and the second thing is just only installing wixtoolset

cevich: No worries, you're timing is fine.  This PR is a "nice to have" at this point, definitely not absolutely required.

I believe I tried both those suggestions with similar results, but I can try moving again.  The other thought I had is: Maybe a reboot is needed?  That would certainly explain how/why it works fine in Podman CI.

cevich:
Force-push: Try installing .net and wixtoolset after a reboot.

cevich:
Note to me: Make sure crun 1.9.3 is present in rawhide and F39 images.

cevich:
After removing workaround, F39 and rawhide are using crun 1.9.2

@edsantiago when you get a chance can you remind me, do we need .3 or is .2 okay?

edsantiago:
.2 is what we have now

cevich:
Thanks Ed.  I'll push this out then (meaning tag it for Renovate to propose updates).  Though Buildah is going to need some special-handling.

cevich:
Hmmm actually, I'm about to go on PTO, so maybe I'll wait to tag.

PR #313, opened 2023-11-16
Try to preserve package versions as artifacts

edsantiago:
package-versions.txt, using rpm or dpkg as needed


lib.sh
```diff
@@ -353,7 +353,8 @@ rh_finalize() {
     # Packaging cache is preserved across builds of container images
     $SUDO rm -f /etc/udev/rules.d/*-persistent-*.rules
     $SUDO touch /.unconfigured  # force firstboot to run
-    common_finalize
+
+    rpm -qa | sort > package-versions.txt
```
edsantiago, line 357: @cevich, very low priority, I'm guessing this is a nested-subnested-VM thing. Is there a quick way to get this file out into where Cirrus can see it?

cevich: Exactly right, with the exception of `cache_images_task` which runs in a container managed by Cirrus.  But that detail may not be important.  I think packer is the thing getting in the way here.  It's responsible for all build-subject VM management, unlike the normal CI flows where Cirrus takes care of this.

Looking at the [packer provisioner docs](https://developer.hashicorp.com/packer/docs/provisioners/file), copying files is a one-way thing :cry: 

I poked around in the cirrus docs quickly, and it doesn't seem there's a way for a VM not running the cirrus-agent process, to interact with cache or artifacts.

Other than bouncing the files through some intermediary, I think the easiest way to exfiltrate a file is through a new cirrus-task.  It's only purpose is to launch the built VMs and use the artifact instruction to scoop up the `package-versions.txt` directly.

Something like (untested):

```yaml
exfiltrate_package_versions_task:
    name: Extract $IMAGE_NAME package-versions
    alias: exfiltrate_package_versions
    gce_instance:
        image_project: "libpod-218412"
        image_name: "${IMAGE_NAME}"
    matrix:
        - env: {"IMAGE_NAME": "fedora-c${IMG_SFX}"}
        - env: {"IMAGE_NAME": "prior-fedora-c${IMG_SFX}"}
        - env: {"IMAGE_NAME": "fedora-netavark-c${IMG_SFX}"}
        - env: {"IMAGE_NAME": "fedora-podman-py-c${IMG_SFX}"}
        - env: {"IMAGE_NAME": "rawhide-c${IMG_SFX}"}
        - env: {"IMAGE_NAME": "debian-c${IMG_SFX}"}
    script: cat /path/to/package-versions.txt  # for debugging
    package_versions_artifact:
        path: /path/to/package-versions.txt
```

cevich: Note: I limited that to only GCE cache images, since generally we assume the AWS images are identical (not 100% correct but likely close enough).

cevich: Hmmm, probably also want to add a dependency in that task, on `cache_images`

edsantiago: Grumble. Sounds annoying (and tedious and risky and time-consuming). Easier to just `rpm -qa` direct to stdout. Would you mind if I did that? I could also `| grep LIST`, but have a moderate preference for not doing so because a few times a year we really want to find the version of X, where X is something-we-did-not-think-about-a-priori. But it's infrequent, so I'm OK with a filtered list too.

Thank you for confirming my guess and for the link to docs.

cevich: > Grumble. Sounds annoying (and tedious and risky and time-consuming).

Maybe not as much as it would seem, at least IMHO.  A matrix like that will run in parallel.  Spinning up a VM takes ~1minute.  Since the task would depend on `cache_images` passing, it wouldn't run at all on a build failure.

The `ccia` container image is able to download cirrus-ci artifacts by regex.  [This is how the name/id table is produced](https://github.com/containers/automation_images/blob/main/.github/workflows/pr_image_id.yml#L72-L74) - downloading all the packer `manifest.json` files and doing a bit a `jq`.

Are there other risks/time/tedium you're concerned about?

edsantiago: Every VM spinup is an opportunity for Cirrus to barf with the "java blah blah" flake.

And, well, there's all that duplication....

edsantiago: >the dreaded java barf [...] Have you been seeing it a lot/more recently?

Oh heck yes, typically a few times a day.

Anyhow, closing, because what I wish to do cannot (apparently) be done cleanly. Thanks for your help.

PR #331, opened 2024-02-26
Bump. Hoping to get rc6 kernel in rawhide

edsantiago:
Signed-off-by: Ed Santiago <santiago@redhat.com>

PR #332, opened 2024-02-28
Support no-clone build-push mode

cevich:
As originally conceived, the build context for each image lives in the respective podman, buildah, and skopeo repositories.  A future set of PRs will move both the source and build automation into the new containers/image_build repository.  This is needed to support images that are point-in-time rebuildable and run test-builds on image context changes.

Add a magic 2nd argument prefix ('.'), and conditionals to prevent cloning the build context repo. This will allow for an interim period where build automation can run from both the current and new repository until the context repos can be moved.


cevich:
@edsantiago FYI I bumped out a bunch of timebomb dates in this PR b/c I only need to make the fedora-based build_push VM (so it can be tested).  Is this going to mess you up in #331 ?

edsantiago:
Can you please merge 331? It's done.

cevich:
Yep, np, done.  BTW, unless you need aa second-look, please feel free to just merge those PRs yourself.  No reason for unnecessary administratia here :grinning:

edsantiago:
I was partway through review. No showstopper concerns.

> No reason for unnecessary administratia here

The no-surprises factor: making sure everyone knows what's going on in this repo. Thanks for merging it.

cevich:
> The no-surprises factor:

Well I appreciate that.  Just don't want you to feel obligated, should the need arise in some more urgent situation.

PR #341, opened 2024-03-27
Update command to install chocolatey packages

l0rd:
Changes `choco install` command to ahdere to [chocolatey scripting best practices](https://docs.chocolatey.org/en-us/choco/commands/#scripting-integration-best-practices-style-guide). 

In particular:
- use `choco upgrade` rather `choco install`
- move the package name to be the first parameter
- explicitly set choco source

### DIFF
--- a/win_images/win-lib.ps1
+++ b/win_images/win-lib.ps1
@@ -35,7 +35,7 @@ function retryInstall {
                 $pkg = @("--version", $Matches.2, $Matches.1)
             }
 
-            choco install -y --allow-downgrade --execution-timeout=300 $pkg
+            choco upgrade $pkg -y --allow-downgrade --execution-timeout=300 --source="'https://community.chocolatey.org/api/v2'"
             if ($LASTEXITCODE -eq 0) {
                 break
             }


cevich (approved):
LGTM.  To get CI un-stuck: run `make IMG_SFX` in the repo root and commit the result.  Think of this similar to a record of the "latest" image tag.

edsantiago:
This isn't actually going to do anything. Builds are working again today. Chocolatey must've just been down yesterday.

cevich:
> I tried to do that but it doesn't work in main:

That's an intentional check Ed added, since pushing would simply fail to build (due to the same thing).  They're put in place to mark temporary workarounds that need to be reviewed at a future date.  Ed's really been doing the best job keeping track of the "why" for each of these timebomb blocks, so I normally would consult with him. But...

> This isn't actually going to do anything. 

...maybe this PR should be closed?  (though we appreciate the time/effort you put into it).

cevich:
> this PR introduces an improvement in the code base

Thanks for pushing back.  I would defer to your expertise on the subject.

@edsantiago since you're already building images, and chocolaty seems to be behaving, would you consider rebasing these changes into your PR?  An alternative is, eat the risk and add the magic `[skip-ci]` onto this PR and just ram it in - I see no compelling reason to subject @l0rd to the complexities of image-building based on the diff here.

win_images/win-lib.ps1
```diff
@@ -35,7 +35,7 @@ function retryInstall {
                 $pkg = @("--version", $Matches.2, $Matches.1)
             }
 
-            choco install -y --allow-downgrade --execution-timeout=300 $pkg
+            choco upgrade $pkg -y --allow-downgrade --execution-timeout=300 --source="'https://community.chocolatey.org/api/v2'"
```
cevich, line 38: (non-blocking) Do you think it might be helpful to other non-windows-expert maintainers, to include a comment here:

```
# ahdere to chocolatey scripting best practices.
# ref: https://docs.chocolatey.org/en-us/choco/commands/#scripting-integration-best-practices-style-guide
```

edsantiago:
I can't fold this into any of my PRs, and I can't condone force-merging it either. Rearranging arg order is fine, but:
* explicit URL is evil. It's like `dnf install --from=today's-fedora-repo.org/today's-api-version`. The URL and/or API version will change in the future, it is not OK to force that onto future maintainers. Install tools should know where to get their packages from
* "upgrade" is not an acceptable term. A normal human being reading that line would see `upgrade` and waste time investigating which version might already be installed, which one we're asking for, and get confused. (It boils down to whether this script is being written for experts-in-podman or experts-in-chocolatey. I believe we are writing for the former).
* finally, yes, any change that's as confusing and hard-to-maintain as this requires a large comment (as you requested).

Since I'm doomed to spend my days rebuilding VMs, I will incorporate two changes into one of my future PRs: rearrange args, and add a link to chocolatey best practices. But I humbly ask you not to incorporate this PR as it stands.

win_images/win-lib.ps1
l0rd: This is a good idea 👍

l0rd:
> I can't fold this into any of my PRs, and I can't condone force-merging it either. Rearranging arg order is fine, but:
> 
>     * explicit URL is evil. It's like `dnf install --from=today's-fedora-repo.org/today's-api-version`. The URL and/or API version will change in the future, it is not OK to force that onto future maintainers. Install tools should know where to get their packages from
> 
>     * "upgrade" is not an acceptable term. A normal human being reading that line would see `upgrade` and waste time investigating which version might already be installed, which one we're asking for, and get confused. (It boils down to whether this script is being written for experts-in-podman or experts-in-chocolatey. I believe we are writing for the former).
> 
>     * finally, yes, any change that's as confusing and hard-to-maintain as this requires a large comment (as you requested).

I agree that the chocolatey scripting "best practices" are counterintuitive and can be hard to maintain. Ignoring them though, can hurt us even more in the future. As per your request I am closing this PR and will let you decide what to include or not in yours.

cevich:
Thanks @l0rd for working with us and sharing your expertise.  I'm confident meaningful improvements will come from this.

PR #349, opened 2024-04-23
Fedora40

cevich:
Depends on https://github.com/containers/automation_images/pull/348

PR #352, opened 2024-05-02
Update to CentOS 9 + misc maintenance

cevich:
***Depends on:*** https://github.com/containers/automation_images/pull/349

* Bump up to CentOS Stream 9 tooling images
* Fix test_build-push_task when skip label present
* Fix wix v4 not installing from chocolaty


edsantiago:
Before I spend much time on this... I'm not sure how this is better than #349? Still a lot of commits, many of them unrelated to the title

cevich:
> Before I spend much time on this

Please don't yet.

> I'm not sure how this is better than 349

It's rebased ontop of it, so the commits will remain cluttered until that PR merges.  Though :thinking: I wonder if it could be independent?

cevich:
Answer: Nope, all the timebomb() explode.

PR #353, opened 2024-05-03
[CI:DOCS] Fix test_build-push failing w/ no_build-push label

cevich:
Previously the build-push task was much more sophisticated and able to run even if a new CI VM image was not produced.  This situation has now changed, and the testing task requires some additional "smarts" to not run when its image wasn't build.

Note: I have to set `[CI:DOCS]` on this otherwise the validate check barfs due to no `IMG_SFX` update.  I cannot update `IMG_SFX` due to a bunch of expired `timebomb()`s.  However, this change (and the `timebomb()`s) was successfully tested when it was originally part of https://github.com/containers/automation_images/pull/349

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -462,7 +462,9 @@ test_ccia_task:
 test_build-push_task:
     name: "Test build-push VM functions"
     alias: test_build-push
-    only_if: *is_pr
+    only_if: |
+        $CIRRUS_PR != '' &&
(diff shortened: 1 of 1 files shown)


cevich:
@edsantiago FYI: I'm just going to merge this as is.  I know it works and took the time to write a (hopefully) helpful commit message :smile:

For some more context: This relates to adding the `no_build-push` label to a PR.  Doing that results in (surprise!) no image being built.  But the testing task didn't know that, so it would fail with a "cirrus: image not found" type error.

PR #355, opened 2024-05-21
Skopeo: Backport a patch to avoid a panic when compiled with Go >= 1.22 

mtrmac:
Try the “easy” way to fix https://github.com/containers/skopeo/pull/2328 .

Warning: I don’t know what I am doing. I’m hoping this PR, even before merging, generates images which can be used to run Skopeo tests to confirm this works as expected.

I have locally tested the `sed` command, but nothing else.

PR #358, opened 2024-05-28
Fix tagging latest after [CI:TOOLING] PR merge

cevich:
After a PR merges a branch-level job runs to tag the new container images.  However, there is a special-case when a magic string is present in the PR title:  No Fedora/Skopeo images were be built, so they should not be tagged be ignored.

Prior to this commit, this special case isn't handled correctly, because `CIRRUS_CHANGE_TITLE` only contains the first-line of the HEAD commit. When executing on a branch, after a PR merge, this would be something like:

`Merge pull request #FOO from some/thing`

Therefore not matching the intended magic string.  Fix this by switching to a check against `CIRRUS_CHANGE_MESSAGE` which includes the entire message.  Importantly, when merged using the github UI, the second line of the commit message should contain the PR description and thus the magic string.

* [Example failure log.](https://cirrus-ci.com/task/5318281827975168)
* [Example failure commit.](https://github.com/containers/automation_images/commit/afe1ced362caf3ad65fc502da6af3567de0266f3)

Note: There's no easy/practical way to test this fix actually works before merging it.  If it fails, the monitoring automation will catch it and send a message to the mailing list.


cevich:
Note: This PR fixes a very minor corner-case issue, I'm not merging it immediately to provide a clear path for #355

ci/tag_latest.sh
```diff
@@ -9,7 +9,7 @@ fi
 
 # This envar is set by the CI system
 # shellcheck disable=SC2154
-if [[ "$CIRRUS_CHANGE_TITLE" =~ .*CI:DOCS.* ]]; then
+if [[ "$CIRRUS_CHANGE_MESSAGE" =~ .*CI:DOCS.* ]]; then
     echo "This script must never run after a [CI:DOCS] PR merge"
```
edsantiago, line 13: Should this be `die`? or `echo + exit`?

cevich: oh ya, good catch.

cevich:
Was going to wait for #355 but it appears that may take a while.  Merging this so the bug is fixed.

PR #359, opened 2024-05-28
Fix missing uuidgen tool

cevich:
Previously this tool was used by a few container images as a half-hearted attempt at thwarting guesses of the credentials filename.  For whatever reason the `uuidgen` command is no longer present in the latest base images but this anti-thwart measure is also unnecessary and not very effective, remove it.


cevich:
Oh, thanks @edsantiago I was intending to merge this, but was stuck in a meeting.

PR #360, opened 2024-05-29
[CI:DOCS] Stop tagging during cron runs

cevich:
Previously the `tag_latest_images` was executing during the daily 'lifecycle' Cirrus-cron job.  This was unintentional, this task should only run after a merge onto the default branch.  Fix the condition.


cevich:
@edsantiago I know you've got a lot in flight right now, and this PR isn't critical.  PTAL when you have a moment and merge as appropriate.  This will help prevent the daily e-mail notices we get about this task failing when it shouldn't even be running.

edsantiago (approved):

PR #364, opened 2024-06-28
[CI:DOCS] Add pre-commit (app) hook to check IMGSFX

cevich:
Intended for use by [the pre-commit
app](https://pre-commit.com/#intro), this hook keeps track of all IMGSFX values pushed, failing when any duplicate is found.  In the case of pushing to PRs that don't build CI VM images, the hook failure must be manually bypassed.  Example `.pre-commit-config.yaml`:

```yaml
---
repos:
  - repo: https://github.com/containers/automation_images.git
    rev: <tag or commit sha>
    hooks:
      - id: check-imgsfx
```


cevich:
Manually tested with:

`pre-commit try-repo . check-imgsfx --verbose --all-files --hook-stage pre-push`

Correctly throws error:
```
Check IMGSFX for accidental reuse........................................Failed
- hook id: check-imgsfx
- duration: 0s
- exit code: 1

FATAL: 20240620t153000z-f40f39d13 has already been used
Please rerun 'make IMG_SFX'
```

If I then `make IMG_SFX` and stage it, running the same command returns:
```
Check IMGSFX for accidental reuse........................................Passed
- hook id: check-imgsfx
- duration: 0s
```

cevich:
force-push: Renamed all `IMGSFX` references to `IMG_SFX`.

cevich:
@edsantiago PTAL, even if you don't use the `pre-commit` app, I'd appreciate a quick look-through for any typos I may have made.

After this merges, my intention is to open a followup with an example `.pre-commit-config.yaml` file specific to this repo.  In addition to your neat IMG_SFX check, it will also validates yaml, EOF, trailing whitespace, shellcheck, and a few more.  But I need a commit-id on main to reference for the new `check-imgsfx`.

check-imgsfx.sh
```diff
@@ -0,0 +1,29 @@
...
+# Ensure CWD is the repo root
+cd $(dirname "${BASH_SOURCE[0]}")
+imgsfx=$(<IMG_SFX)
+
+if [[ -z "$HOME" ]]; then
+    echo "FATAL: $$HOME is undefined" >&2
+    exit 1
+fi
+
+# Pre-commit runs in a temp. dir, history must persist via absolute path.
```
edsantiago, line 17: "runs in a temp dir" -- is that really true? I can't find it in the pre-commit docs. It strikes me as preposterous if true: the entire point of a pre-commit hook is to run tests in the repo, isn't it?

Anyhow, the `cd` on line 9 negates this. You can still keep the history file under `$HOME`, but it's not really a "must". I have a slight preference for keeping it in the same repo dir (possibly with a `.gitignore` entry) but won't block over it.

edsantiago (commented):
pre-commit looks intriguing, thanks for the pointer. I can't verify the config. Some feedback on the sh part.

check-imgsfx.sh
cevich: > runs in a temp dir

I couldn't tell if the actual runs happen in a temp. local clone, but certainly the `pre-commit try-repo ...` testing did.  It showed the tmp dir output when I ran it verbosely.  Worse, it would run the script from there as well.  I can try rigging up a `.pre-commit-config.yaml` that points at the local repo. and see if the `pre-push` can use an in-repo dir.

Another thing this comment should probably say:  The `~/.cache/pre-commit` dir. is created/used by the tool itself for logging and whatnot.  So it's not as if I just picked some random location :wink:

check-imgsfx.sh
```diff
@@ -0,0 +1,29 @@
...
+# 2024-06-28 cevich
+
+set -eo pipefail
+
+# Ensure CWD is the repo root
+cd $(dirname "${BASH_SOURCE[0]}")
+imgsfx=$(<IMG_SFX)
+
+if [[ -z "$HOME" ]]; then
+    echo "FATAL: $$HOME is undefined" >&2
```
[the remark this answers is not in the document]
cevich: Oh woops, that's what I intended.

cevich:
Force-push: Remove use of `$HOME` based storage of history.  Add intended use/purpose comment block to script.

@edsantiago this is ready for final review/merging.  I'll start working on a followup PR to add an example pre-commit configuration file.

check-imgsfx.sh
edsantiago: Comment is now wrong two ways :-)

check-imgsfx.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+#
+# 2024-01-25 esm
+# 2024-06-28 cevich
+#
+# This script is intended to be used by the `pre-commit` utility, or it may
+# be manually copied (or symlined) as local `.git/hooks/pre-push` file.
```
edsantiago, line 7: typo, symlinked

check-imgsfx.sh
cevich: Oh woops did I leave that in there, sorry :relaxed:

PR #370, opened 2024-07-05
[CI:TOOLING] Increase obsolete image flagging and pruning

cevich:
Increase obsolete image flagging and pruning

It was observed in the Cirrus-CI cron logs, that only the total
number of images scanned is reported.  Fix this by giving more
useful info., like the number of candidates for obsolete/pruning.

Related, the original restriction of `10` obsolete/prune images
was originally put in place when only a few repos utilized Cirrus-CI
VMs and image building was substantially more infrequent.  The
reason it exists is to prevent potential catastrophe should the `meta`
time stamp updating tasks have a bug or some other related failure occur.
Increase the limit to `50` so deletions may proceed much more rapidly.

*Note:* "Obsolete" images still live w/in a 30-day window where they can
be recovered if need be.  It's simply that any attempted use by CI will
fail, putting someone on notice that image recovery may be necessary.


edsantiago:
LGTM

cevich:
Force-push: No code changes, just commit/message cleanup.

PR #373, opened 2024-08-08
new images

Luap99:
Mainly to pull in a new pasta version to fix
https://github.com/containers/podman/issues/23517


Luap99:
passt 2024-07-26 in fedora-aws is not good. Are we not using the same repos there or do the aws images have some older mirrors?

edsantiago:
e.g., cd783f07c3bf171f655f536cfb668275ff6e8d79 (that's a remove, not an add, but you get the idea)

cevich:
> Yeah I figure they use their own aws specific mirrors

That sounds plausible to me, in GCP we're pulling in [the generic cloud image](https://github.com/containers/automation_images/blob/83c9b1661c41f6e8d1daab5c4d6774083c5cb85e/get_fedora_url.sh) and [molding it into a GCP-compatible image via a nested-virt builder](https://github.com/containers/automation_images/blob/83c9b1661c41f6e8d1daab5c4d6774083c5cb85e/base_images/cloud.yml#L65-L69).  For AWS, we [search AWS-specific images build and published by the fedora project](https://github.com/containers/automation_images/blob/83c9b1661c41f6e8d1daab5c4d6774083c5cb85e/base_images/cloud.yml#L117-L124) so they easily could have AWS-specific content, like repo-mirror configuration.  It should be easy to observe this using a `hack/get_ci_vm.sh` somewhere, against an EC2-based task.

PR #374, opened 2024-08-09
Reduce impact of networking slowdowns

cevich:
Previously if a repository server, the internet, or the execution
environment experienced some kind of networking slowdown, it could lead
to a package install or update timeout failure.  Increase resiliency in
these situations with additional retries, timeouts, and lowered minimum
rates.  Also increase the timeout on the related Cirrus-CI tasks.


cevich:
@edsantiago PTAL when you have a moment, this isn't critical.

cevich:
> do you have any examples of when either of these has failed

I don't, this is purely speculative.  I do recall on occasion, an entire repo. mirror (or single package) will flake irrecoverably.  In those cases you just need to wait and try again later.  Sometimes the next day.

> that jobs flake because they run into the Cirrus timeout

That's my recollection as well, and yes I probably should increase that timeout as well.

cevich:
force-push: Add/increase timeouts for the Cirrus-CI tasks.

@edsantiago In case it's not clear, the primary target is the nested-virt Fedora base image builds.  But I'm not married to this PR, it's really just speculation on increasing build-problem/flake resiliency.  If you feel it's unnecessary, adds too much complexity, or whatever I'll just close it w/o any hard feelings.

edsantiago (approved):
LGTM

edsantiago:
Identical to https://github.com/containers/automation_images/pull/375#issuecomment-2284446891

PR #375, opened 2024-08-09
Catch Fedora-base image update problems early

cevich:
Previously updates were disabled due to the cloud VM only having 2-gig and the nested-VM only having 1-gig of memory.  Allow Fedora base-image package updates by increasing the available resources.  Enabling base-level (esp. kernel) package updates early supports spotting fundamental image problems early.  Otherwise they may not be found until a set of images is deployed downstream.

Also, update a few comments relating to followup package update.

PR #380, opened 2024-08-16
[CI:TOOLING] Track image IDs instead of tar exports

cevich:
Previously all container builds run by the Makefile managed them based
on presence/absence of a docker-archive tar file.  Producing these
exports is time-consuming and ultimately unnecessary extra work.  The
tar files are never actually consumed in a meaningful way by any other
targets.  Further, most of the container builds in CI run in parallel,
simply throwing away the tar when finished.

Fix this by switching to management based on image-ID files instead.
The only exception is the `imgts` image and images which are based on
it.  For those, some special handling is required (already done by the
CI build script), so some comments were added to assist.

Also, remove the `bench_stuff` target entirely as this has long since
been retired.


cevich:
I think this is ready for review.  Container image build times with [these changes](https://github.com/containers/automation_images/commit/12b728a6a1f84932e7fb7eae264129a75b9bd169):

imgts 06:47
ccia 2:23
gcsupld 1:21
get_ci_vm 1:21
imgobsolete 1:21
imgprune 1:21
orphanvs 1:21

Compared to [a past build](https://github.com/containers/automation_images/pull/375):

IMGTS 08:13
ccia 02:21
gcsupld 09:59
get_ci_vm 01:43 
imgobsolete 09:36
imgprune 09:32
orphanvms 09:46

Likely the biggest build-time boost is the result of pushing/pulling the imgts, as it's used as a base-image for most of the others.

cevich:
Ed and/or Paul, if you have time would you mind giving this a look-through?

Makefile
```diff
@@ -150,8 +153,14 @@ timebomb-check:
 	        false; \
 	    fi
 
+# Given the path to a file containing 'sha245:<image id>' return <image id>
```
edsantiago, line 156: Not worth a repush on its own

Makefile
```diff
@@ -163,19 +172,18 @@ ci_debug: $(_TEMPDIR)/ci_debug.tar ## Build and enter container for local develo
 		-e GAC_FILEPATH=$(GAC_FILEPATH) \
 		-e AWS_SHARED_CREDENTIALS_FILE=$(AWS_SHARED_CREDENTIALS_FILE) \
 		-e TEMPDIR=$(_TEMPDIR) \
-		docker-archive:$<
+		$(call imageid,$<) $(if $(DBG_TEST_CMD),$(DBG_TEST_CMD),)
```
edsantiago, line 175: I don't understand this. Does this magic image have an entrypoint or command defined for when it's invoked without args? If so, did you confirm that the `true` here actually overrides the default? Because sometimes you can override via command-line, and sometimes you need to specify `--entrypoint`. (I could figure this out myself given enough time but my brain hurts from this). Anyhow, CI got skipped for this PR so I can't verify tests.

edsantiago (approved):
This is completely over my head, but LGTM in theory with one question

Makefile
cevich: Oh whups.  I can fix that NP, this PR builds quickly.

Makefile
cevich: This is intended for interactive-use by humans, and should probably be documented in the README.md.  The entrypoint is simply the default 'bash'.  It's for developing/debugging the dockerfile-based Cirrus-CI tasks, since the image-build happens w/in Cirrus (i.e. it's never pushed anywhere):

(.cirrus.yml)
```yaml
...cut...
imgts_build_task:
    ...cut...
    container:
        dockerfile: "image_builder/Containerfile"
    ...cut...
...cut...
```

> Anyhow, CI got skipped

Careful, CI wasn't entirely skipped.  The builds and new test tasks did run :wink: 

I'll add some quick docs for this.

---

FWIW, the `image_builder_debug` target (also intended for humans) is [already documented](https://github.com/containers/automation_images/blob/main/README.md#debugging--locally-driving-vm-image-production).

Luap99 (commented):
I don't feel qualified to review this. There is a lot of makefile logic that I cannot follow right now. Anyhow using image ids to pass the images around to other tasks seems better

cevich:
Thanks guys.  I understand `make` was probably not the best choice originally.  But I'm hopeful these simplifications, speedups and doc updates are at least a small improvement.

PR #376, opened 2024-08-14
Remove unused packages and update python versions

inknos:
python-xdg was removed as a dependency
https://github.com/containers/podman-py/commit/8d1020ecc88c55890ce0e1fa843803247911df0b

tests are currently done for py12
https://github.com/containers/podman-py/commit/330cf0b5011a114d0bc4e86477e4e5c868dbad31

PR #388, opened 2024-10-15
Try zstd:chunked in local cache registry

edsantiago:
There's a strong push to use zstd:chunked by default, but
very little testing of it. This forces zstd:chunked on
images in the local mirror cache.

Also, add new testimage:20241011 (podman#24238)


edsantiago:
- 20241015t143000z-f40f39d13
- 20241015t085508z-f40f39d13 &#x21D1; (built in #376)

|  |*debian*|*prior-fedora*|*fedora*|*fedora-aws*|*rawhide*|
|----|----|----|----|----|----|
|*podman*       | 5.2.3+ds1-4     | 4.9.4-1         | 5.2.4-1         | 5.2.3-1         | 5.2.4-1         |
|               | 5.2.3+ds1-3&nbsp;&#x21D1; |                |                |                |                |

Luap99 (commented):
Do I need to get popcorn ready to see if this passes podman CI?

cache_images/local-cache-registry
```diff
@@ -82,6 +82,7 @@ declare -a IMAGELIST=(
     testimage:00000004
     testimage:20221018
     testimage:20240123
+    testimage:20241011
```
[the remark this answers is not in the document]
edsantiago: I have a `find-unused-images` script but I don't quite trust it yet. Am planning to fine-tune it then submit it as part of the repo one day. It's just a oneliner:
```bash
for i in $(sed -ne '/IMAGELIST=/,/^[^ ]/p' <cache_images/local-cache-registry | sed -ne 's/^  *//p');do grep -q -R $i ../libpod/test ../buildah/tests || echo "unused $i";done
```
...and its current output is:
```
unused fedora/python-311:latest
unused podman_python:latest
unused testimage:20240123       <<< yay nice catch
unused testimage:multiimage     <<< wrong and hard to test for and maybe I can find a way to not need it
unused podman/stable:v4.3.1     <<< obviously wrong, but hard to check for
unused podman/stable:v4.8.0     <<< ibid
```
The python ones I think are used in libpod compose tests, but have never taken the time to investigate why

edsantiago:
@Luap99 low-priority question: should I add `nv-rust:1.76` to the cache list? On the assumption that some future day netavark will be able to bump VMs?

Luap99:
> @Luap99 low-priority question: should I add `nv-rust:1.76` to the cache list? On the assumption that some future day netavark will be able to bump VMs?

No we do not use this image in our CI images at all, AFAIK all the cirrus container tasks run in their community cluster so it will be pulled there automatically

edsantiago:
- 20241017t115000z-f40f39d13
- 20241016t144444z-f40f39d13 &#x21D1; (built in #389)

|  |*debian*|*prior-fedora*|*fedora*|*fedora-aws*|*rawhide*|
|----|----|----|----|----|----|
|*containers-common*| ?               | 1-99            | 0.60.4-2        | 0.60.4-1        | 0.60.4-3        |
|               |                |                | 0.60.4-1&nbsp;&#x21D1; |                | 0.60.4-1&nbsp;&#x21D1; |

cache_images/local-cache-registry
```diff
@@ -219,6 +220,7 @@ function cache_image() {
     for retry in 1 2 3 0;do
         skopeo --registries-conf /dev/null \
                copy --all --dest-tls-verify=false \
+               --dest-compress-format zstd:chunked \
```
mtrmac, line 223: This changes image digests, which are hard-coded in tests (e.g. `ALPINELISTDIGEST`).

That’s fine in principle — and I’ll update the test digests in the integration PR — I’m just noting that it’s expected that some tests will start failing with this.

---

Outside of the current zstd:chunked testing effort, consider using `--preserve-digests` in the command to make any accidental digest changes (e.g. if an unusual image weren’t supported by the cache registry and triggered a format conversion) an error, to make sure this creates a bit-for-bit-identical copy.

edsantiago:
- 20241022t120000z-f40f39d13
- 20241016t144444z-f40f39d13 &#x21D1; (built in #389)

|  |*debian*|*prior-fedora*|*fedora*|*fedora-aws*|*rawhide*|
|----|----|----|----|----|----|
|*kernel*       | 6.11.4-1        | 6.5.6-300       | 6.8.5-301       | 6.8.5-301       | 6.8.5-301       |
|               | 6.11.2-1&nbsp;&#x21D1; |                |                |                |                |
|*grub2-common* | 2.12-5          | 2.06-121        | 2.06-123        | 2.06-123        | 2.12-10         |
|               |                |                |                |                | 2.12-9&nbsp;&#x21D1; |
|*buildah*      | 1.37.5+ds1-1    | 1.37.1-1        | 1.37.3-1        | 1.37.3-1        | 1.37.5-1        |
|               | 1.37.4+ds1-1&nbsp;&#x21D1; |                | 1.37.4-1&nbsp;&#x21D1; |                | 1.37.4-1&nbsp;&#x21D1; |
|*containers-common*| ?               | 1-99            | 0.60.4-2        | 0.60.4-2        | 0.60.4-5        |
|               |                |                | 0.60.4-1&nbsp;&#x21D1; | 0.60.4-1&nbsp;&#x21D1; | 0.60.4-1&nbsp;&#x21D1; |
|*criu*         | 3.17.1-3        | 4.0-1           | 4.0-1           | 4.0-1           | 4.0-2           |
|               |                |                |                |                | 4.0-1&nbsp;&#x21D1; |
|*podman*       | 5.2.4+ds1-1+b1  | 4.9.4-1         | 5.2.3-1         | 5.2.3-1         | 5.2.5-1         |
|               | 5.2.3+ds1-4&nbsp;&#x21D1; |                | 5.2.4-1&nbsp;&#x21D1; |                | 5.2.4-1&nbsp;&#x21D1; |
|*runc*         | 1.1.12+ds1-5.1+b1 | 1.1.12-1        | 1.1.12-3        | 1.1.12-3        | 1.1.12-4        |
|               | 1.1.12+ds1-5+b1&nbsp;&#x21D1; |                |                |                |                |
|*skopeo*       | 1.16.1+ds1-1+b1 | 1.16.1-1        | 1.16.1-1        | 1.16.1-1        | 1.16.1-2        |
|               | 1.16.1+ds1-1&nbsp;&#x21D1; |                |                |                |                |
|*systemd*      | 256.7-2         | 254.18-1        | 255.13-1        | 255.13-1        | 256.7-1         |
|               | 256.7-1&nbsp;&#x21D1; |                |                |                |                |

cache_images/local-cache-registry
edsantiago: Thanks for this. The new images built just now use `--preserve-digests`. Would you care to try with `c20241022t120000z-f40f39d13` ?

mtrmac: I’m sorry, with `--preserve-digest` + `--dest-compress-format`, `--dest-compress-format` is, AFAICS, ignored (because that requires a digest change). I’ll inspect the images to confirm.

Arguably it’s a bug that this option combination is not rejected.

edsantiago: Ack, thanks, I misunderstood.

So, back to your initial comment: "This changes image digests". Are these changes reliably consistent? When we make new VM images next week/year, different version of skopeo, different kernel/filesystem/tools, different phase of the moon, will the image digests be the same?

If so, then it might be okay to change the digests in the tests, **but** it's important to remember that e2e tests need to be able to run using local cache and also run using quay. Tests can determine the environment via `UsingCacheRegistry()` (returns bool). This is ugly but it's a potential way to get tests to pass.

If digests are not consistently reproducible, all bets are off and the tests will need to be completely reimagined.

mtrmac: > Are these changes reliably consistent?

No; there is no guarantee that compression produces consistent results even with one version, and the output can (and, from time to time, does) change when the implementation changes.

For the purpose of the current experiment, I’ll hard-code the digests as they exist in the local cache. But, other than that, we’ll indeed need to specifically maintain zstd:chunked images on quay.io — or, alternatively, not hard-code the expected digests in tests, and obtain them at runtime, or maybe at the cache build time, using something like Skopeo.

Luap99: I think the digest always depends on the compression, so as long as the compression is reproducible it should produce the same digest?

In any case is there any need to hard code the digests? Skopeo inspect can give us that
```
skopeo inspect --no-tags  --format {{.Digest}} docker://quay.io/libpod/testimage:20241011
```
(The command seems quite slow for some reason, so it isn't something I would like to run in every test but only once and cache it)

mtrmac: Compression isn’t reproducible.

Using `skopeo inspect` is definitely an option — noting that it decreases the test coverage a bit: if c/image were broken, we’d see consistent incorrect data from both Skopeo and the Podman binary being tested.

mtrmac: Sure — and the risk is not that high.

---

Maybe, first, we need some thought about the final test plan: do we actually want to add a gzip / zstd:chunked _dimension_ to the testing matrix, to make this a variable? Maybe the necessary testing coverage can be achieved by a few dedicated tests/images. That’s very unclear to me.

edsantiago: Exactly. This is one of the questions on my radar, that I don't yet know enough to answer. My hope is that we can do an integration, mixed set of images, but as of this writing my recollection is that some images have worked (with zstd:chunked) and some have not, and that makes me uncomfortable. It's possible that the reason for this is that some _tests_ use different images and that's what triggers the failures, but I haven't spent the time to investigate. Thank you for bringing this up.

mtrmac:
zstd:chunked now has some specific tests, e.g. https://github.com/containers/podman/pull/24413 ; https://github.com/containers/podman/pull/25007 . Those are fairly narrowly targeted, and don’t require the somewhat disruptive conversion of _all_ existing test images to zstd.

PR #397, opened 2024-12-18
Add 7zip Windows compression utility

baude:
The Fedora images used to test libhvee are now being shipped with xz compression.  Because the golang xz decompression is extremely slow, I'm proposing to use this command line utility.


baude:
Do not merge or delete until further notice.

PR #398, opened 2025-01-06
new images

Luap99:
Removed two timebombs that no longer apply, composefs is installed in the main package list and the pasta version is in stable now.


Luap99:
```
    debian: The following packages have unmet dependencies:
    debian:  git-daemon-run : Depends: git (< 1:2.39.5-.) but 1:2.47.1-1 is to be installed
    debian: E: Unable to correct problems, you have held broken packages.
    debian:     exit(100)
    debian: Retry attempts exhausted
```

And the fun begins... Not even getting images build this time :/

PR #403, opened 2025-03-27
Fedora image: install koji and fedora-distro-aliases

lsm5:


base_images/fedora_base-setup.sh
```diff
@@ -17,7 +17,7 @@ REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
 source "$REPO_DIRPATH/lib.sh"
 
 declare -a PKGS
-PKGS=(rng-tools git coreutils cloud-init)
+PKGS=(rng-tools git coreutils cloud-init koji python3-fedora-distro-aliases python3-koji-cli-plugins)
```
Luap99, line 20: this is the wrong place to add packages, we don't use the base images, we add all the packages in the cache images here
https://github.com/containers/automation_images/blob/564840b6bccbcc889b97755ca5c5393d7bbf67e3/cache_images/fedora_packaging.sh#L40

Luap99:
https://github.com/containers/automation_images/pull/404

PR #408, opened 2025-06-27
remove podman-py

Luap99:
It only uses tmt now and not cirrus anymore. So delete all the image build infra for it.


Luap99:
cc @inknos @timcoding1988

inknos:
LGTM

PR #410, opened 2025-07-18
skopeo_cidev: Depend on docker-distribution

lsm5:
Having the registry binary named `registry-v2` causes trouble for `make test-integration-local`. The registry binary provided by the docker-distribution package is just `/usr/bin/registry`.

So, depending on docker-distribution should make things simpler, more consistent and usable regardles of CI / testing environment.


skopeo_cidev/setup.sh
```diff
@@ -9,7 +9,6 @@ set -e
 declare -a req_vars
 req_vars=(\
     REG_REPO
-    REG_COMMIT
```
mtrmac, line 12: Also delete the definition from `skopeo_cidev/Containerfile`?

mtrmac (commented):
For the record, this means that instead of testing interoperability with a specific frozen version, we will be testing interoperability with the latest one — and that package updates can therefore trigger new test failures.

That is probably an improvement — I just want this mentioned.

skopeo_cidev/setup.sh
lsm5: yup, it's removed, see the Containerfile diff. Unless you meant something else.

mtrmac: Yes, that’s what I meant.

mtrmac (approved):
Thanks!

PR #411, opened 2025-07-29
WIP: Install podman-sequoia in rawhide images

mtrmac:


Makefile
```diff
@@ -132,13 +132,13 @@ help: ## Default target, parses special in-line comments as documentation.
 # names and a max-length of 63.
 .PHONY: IMG_SFX
 IMG_SFX:  timebomb-check ## Generate a new date-based image suffix, store in the file IMG_SFX
-	$(file >$@,$(shell date --utc +%Y%m%dt%H%M%Sz)-f$(FEDORA_RELEASE)f$(PRIOR_FEDORA_RELEASE)d$(subst .,,$(DEBIAN_RELEASE)))
+	$(file >$@,$(shell date -u +%Y%m%dt%H%M%Sz)-f$(FEDORA_RELEASE)f$(PRIOR_FEDORA_RELEASE)d$(subst .,,$(DEBIAN_RELEASE)))
```
Luap99, line 135: that change seems unrelated?

cache_images/fedora_packaging.sh
```diff
@@ -174,6 +174,10 @@ if [[ "$PACKER_BUILD_NAME" =~ fedora ]]; then
         python3-requests
         python3-requests-mock
     )
+else # podman-sequoia is only available in Rawhide
+    INSTALL_PACKAGES+=( \
+        podman-sequoia
+    )
```
Luap99, line 180: here it would be good to add a timebomb as reminder to add sequoia to the older versions as well once available. I guess if it is only rawhide for now it should be set to November once f43 is released at least.

Makefile
mtrmac: Oops, yes. It was supposed to help with macOS compatibility, but it isn’t enough, I’ll drop it.

mtrmac:
From https://github.com/containers/podman/pull/26674 :
> these images are currently unusable for podman CI to the useradd subuid thing

I’ll return to this in a few days.

Luap99:
According to https://bodhi.fedoraproject.org/updates/FEDORA-2025-3b1a650972 shadow-utils should work again and assign subuids.

Makefile
mtrmac: I have fixed the rule on macOS, and turned that into a separate commit.

timcoding1988: i have the same trouble when using this in macOS

PR #412, opened 2025-08-01
Windows: choco install make

lsm5:
Needed to run validate task on libhvee. The current TMT job to run validate fails because it's on Fedora 42, while the same task passes on Cirrus with Windows.

Ref: https://github.com/containers/libhvee/pull/242


baude:
LGTM

win_images/win_packaging.ps1
```diff
@@ -33,4 +33,7 @@ Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V-Management-
 # Install WSL, and capture text output which is not normally visible
 $x = wsl --install; Check-Exit 0 1 # wsl returns 1 on reboot required
 Write-Host $x
+
+# Install make (required by libhvee)
+choco install make
```
Luap99, line 38: This package must be added to line 12 for the "proper" install way.

Luap99 (changes requested):
I thought we don't want to use make on windows? Regardless a package more or less I don't mind but it should be added in the proper install place.

PR #414, opened 2025-08-19
Set up TLS for the local-registry service

nalind:
Set up a self-signed TLS certificate for the local-registry service.


Luap99:
I picked your commit into https://github.com/containers/automation_images/pull/415 as I also need some change here and realistically we can only have one build out of a PR at any given time without causing conflicts.

nalind:
Sounds good, closing this one.

PR #416, opened 2025-09-04
build new images

Luap99:


Luap99:
Images appear to be broken in a weird way, VM is created in GCE but in cirrus it stays in scheduled state forever?!
In the serial console I cannot see any errors logged, I guess I wait until next week and try again

mtrmac:
@Luap99 I suppose I should hold off on making any changes in this repo for now?

(Motivation: https://github.com/containers/skopeo/pull/2700#issuecomment-3271323602 means I’d like to update `rust-podman-sequoia`, and to include it in a non-Rawhide image).

mtrmac:
Oops, I should be paying more attention to Fedora, I didn’t realize F43 is not out yet.

PR #417, opened 2025-10-02
create new automation VM image 10-30-25

timcoding1988:
new automation image


Luap99:
Note please do not remove these work around by hand, I do them in separate git commits intentionally for proper history tracking. So please do use `git revert -s` so when going through the git log it is easy to find when they we removed again.

But anyhow seems like we still need the debian kernel work around and likely have to wait for 6.17 until we can revert it. So just bump the timebomb date until November or so.

PR #418, opened 2025-10-06
Windows image: use WiX version 5.0.2 

l0rd:
This PR points the WiX toolset version to 5.0.2 to avoid using version 6 when building the installer

### DIFF
--- a/win_images/win_packaging.ps1
+++ b/win_images/win_packaging.ps1
@@ -22,8 +22,9 @@ Invoke-WebRequest -Uri https://dotnet.microsoft.com/download/dotnet/scripts/v1/d
 # Configure NuGet sources for dotnet to fetch wix (and other packages) from
 & 'C:\Program Files\dotnet\dotnet.exe' nuget add source https://api.nuget.org/v3/index.json -n nuget.org
 
-# Install wix
-& 'C:\Program Files\dotnet\dotnet.exe' tool install --global wix
+# Install wix. Version should match the one in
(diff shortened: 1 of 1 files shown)


l0rd:
@Luap99 PTAL

win_images/win_packaging.ps1
```diff
@@ -33,4 +34,15 @@ Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V-Management-
...
+
+# Add a WSL configuration file
+# The `kernelBootTimeout` configuration is to prevent CI/CD flakes
+# See https://github.com/microsoft/WSL/issues/13301#issuecomment-3367452109
+$wslConfigPath = "$env:UserProfile\.wslconfig"
+$wslConfigContent = @"
+[wsl2]
+kernelBootTimeout=300000 # 5 minutes
+"@
+Set-Content -Path $wslConfigPath -Value $wslConfigContent
```
Luap99, line 46: This is per user right due UserProfile?

I am not sure about the windows build process at all but I am not sure if this here then runs as the same user as our podman tests. Would it be easier to configure in the podman setup instead where we know we run under the correct user https://github.com/containers/podman/blob/main/contrib/cirrus/win-podman-machine-test.ps1?

Luap99:
Thanks @l0rd

@timcoding1988 can you cherry-pick this commit into your PR https://github.com/containers/automation_images/pull/417 since we can only have "one image build" (PR) active at a time anyway so best to fold this into your existing build PR.

PR #430, opened 2025-11-22
Podman6: Remove slirp4netns

lsm5:


Luap99:
new automation repo for podman is in https://github.com/podman-container-tools/automation, this repo is dead now as we no longer use cirrus

PR #431, opened 2025-12-12
Add new users in Windows image

l0rd:
Related to https://github.com/containers/podman/issues/27597


Luap99 (commented):
is there a reason why we would need to create the users as part of the image?

The way we have been doing things on the linux side is that the rootless users are created dynamically in the ci setup script in the podman repo. The big advantage being of course it doesn't depend on the image building process and you can create "random" names so we know the code has now the code or image has no special behavior regarding a certain username.

Luap99:
new automation repo for podman is in https://github.com/podman-container-tools/automation, this repo is dead now as we no longer use cirrus.

We do not have any windows specific logic now, we just use the default github runners so please check if they work for your use case

PR #438, opened 2026-03-19
install netavark v2 from upstream

Luap99:
We need the new netavark code to run container-libs/podman tests on it.


Luap99:
untested of course so far but I think that should unblock the CI side once it works 

cc @timcoding1988  @mheon @ashley-cui @Honny1

Honny1 (commented):
LGTM, as far as I can tell. 

Just to be sure, do we need to set an SELinux label for that binary?

timcoding1988:
LGTM

Luap99:
> Just to be sure, do we need to set an SELinux label for that binary?

I don't think is netavark/aardvark-dns do not have specific polices set and on my local system I always copy the binary without selinux and never had problems. Normal copy seems to result in `unconfined_u:object_r:usr_t:s0` which I think is fine.

lib.sh
```diff
@@ -286,6 +286,25 @@ unmanaged-devices=interface-name:*podman*;interface-name:veth*
 EOF
 }
 
+# For Podman 6 we need to test the upstream main code with netavark v2 which is not yet packaged.
```
mtrmac, line 289: Should this have a time bomb?

mtrmac (commented):
I know nothing about the netavark versions/features, the implementation looks plausible.

lib.sh
Luap99: this would certainly make sense though not worth repushing/rebuilding for so I file an issue here to remember to remove it

Luap99: https://github.com/containers/automation_images/issues/439

PR #440, opened 2026-04-15
New images 2026-04-15

ashley-cui:
Grabbing new netavark from upstream


Luap99:
Note if CI passes here the bot adds a comment with the image ID which you then can use in container-libs and podman PR.

The PR here does not need to be merged for that. Only once we know this works in Podman/container-libs we merge here.

Luap99:
Oh an expect CI to take like 2 hours here, the simplified Readme says one but it seems to take much longer these days.

ashley-cui:
all good, i'll open the PR on container-libs and podman once the comment is posted!

Luap99:
The container-libs PR is merged so I merge this even if there are some podman test failures

PR #445, opened 2026-05-08
remove udica project

Luap99:
I requested a removal of the GCP project internally since cirrus CI will shut down and we do not maintain udica.

To ensure the automation keeps working for now until June drop the project from this list as suggested by Chris.

### DIFF
--- a/gcpprojects.txt
+++ b/gcpprojects.txt
@@ -10,4 +10,3 @@ netavark-2021
 oci-seccomp-bpf-hook
 skopeo
 storage-240716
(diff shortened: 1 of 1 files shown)


Luap99:
cc @cevich

cevich:
I think this is fine to force-merge.  IIRC the automation that uses this file is https://github.com/containers/automation_images/blob/main/.github/workflows/orphan_vms.yml  I can't be sure, but that might even be the ONLY place it's used (other than humans looking at it).

